PR #8, opened 2019-09-19
Django DRF Prototype Version

BethanyG:
Version of Django-concept using Django DRF and this working spec:  https://app.swaggerhub.com/apis-docs/billglover/CodeBuddies/0.0.1#/resources/getResources

The model uses a JSON field for the tags.  UUIDs are made up.  Am mocking data for API testing, and will add SQL/CSV for import into as soon as it's done.

Please note that the MODEL is named "Resource", but the APP is named "Resources".


BethanyG:
The `tags` field is JSON.  This is the current format:

```
{
	"tags": [{
			"id": "20fba58a-2702-4378-b0f9-5ac0193d4965",
			"name": "Django"
		},
		{
			"id": "32cd4ddc-6a8e-48b2-be60-45de1784e5c7",
			"name": "Python"
		}, {
			"id": "f81f08af-6bdd-465f-8aba-1693e7348f7d",
			"name": "REST"
		}
	]
}
```

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
```diff
@@ -1,3 +1,47 @@
...
+    date_published = models.DateTimeField(default=timezone.now)
+
+    # creation date of resource entry
+    created = models.DateTimeField(auto_now_add=True)
+
+    # modification date of resource entry
+    modified = models.DateTimeField(default=timezone.now)
+
+    # [ video, podcast, talk, tutorial, course, book, blog ]
+    type = models.CharField(max_length=100)
```
brylie: These options can be constrained with a `CHOICES` argument.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
```diff
@@ -1,3 +1,47 @@
...
+
+    paid = models.BooleanField(null=True)
+
+    # JSON here, so we can parse them easier....
+    tags = pg.JSONField()
```
brylie: I think tags should be a `ManyToMany` relationship, to a `Tags` table. Serialization/deserialization is relatively easy with DRF.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
```diff
@@ -1,3 +1,47 @@
+import uuid
+import django.contrib.postgres.fields as pg
 from django.db import models
+from django.utils import timezone
 
-# Create your models here.
+class Resource(models.Model):
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid1, editable=False)
```
brylie: This field is implicitly defined by Django, so does not need to be part of our model definition.

cbv3_django_prototype/cbv3_django_prototype/resources/urls.py
```diff
@@ -0,0 +1,12 @@
...
+from rest_framework import routers
+from . import views
+
+router = routers.DefaultRouter()
+router.register(r'resources', views.ResourceListView, basename='resources')
+
+urlpatterns = [
+    path('', include(router.urls)),
+    path('api-auth/', include('rest_framework.urls', namespace='rest_framework'))
+    #path('<int:id>/', views.DetailResource.as_view()),
```
brylie: Lets remove commented code. We can add it when it is relevant.

cbv3_django_prototype/cbv3_django_prototype/resources/views.py
```diff
@@ -1,3 +1,14 @@
...
+from .serializers import ResourceSerializer
+from rest_framework import filters, viewsets
+from rest_framework.response import Response
+
+
+class ResourceListView(viewsets.ModelViewSet):
+    queryset = Resource.objects.all()
+    serializer_class = ResourceSerializer
+    filter_backends = [filters.SearchFilter]
+    search_fields = ['id', 'tags']
```
brylie: We probably want to also search by `title` and `description`.

cbv3_django_prototype/config/settings/base.py
```diff
@@ -42,9 +42,21 @@
 # ------------------------------------------------------------------------------
 # https://docs.djangoproject.com/en/dev/ref/settings/#databases
 
+#DATABASES = {
+#    "default": env.db("DATABASE_URL", default="postgres:///cbv3_django_prototype")
+#}
+
 DATABASES = {
-    "default": env.db("DATABASE_URL", default="postgres:///cbv3_django_prototype")
+    'default': {
```
brylie: Lets add instructions to the README about setting up the Postgres Docker container for this to work.

cbv3_django_prototype/config/settings/base.py
```diff
@@ -42,9 +42,21 @@
...
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'cbv3_django_prototype',
+        'USER': 'postgres',
```
brylie: Lets use environment variables for database credentials and path.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
brylie: Disregard. I ser that we want to use UUID rather than sewuential IDs. ☺

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
brylie: The idea is to use relational capabilities of the database first, before resorting to storing nested data. JSON should be used sparingly, if at all, and for specific optimixatiins, like when joins might be too slow.

Using a proper ManyToMany approach here might have other benefits like allowing for autocomplete or clustering related resources (like a recommendation engine).

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
lpatmo: ++ Those options in the comment look pretty comprehensive to me, fwiw.

cbv3_django_prototype/config/settings/base.py
lpatmo: Thanks for the suggestion. I remember this was a pain (IIRC Bethany actually helped me through this during the first/only hangout we've done for this repo at https://zoom.us/recording/play/Ttk6GpLLMJunjTjWpH577Q66cmnvs1dTGf9OCo2CzdwX45NGFFSUfzoIumAnMV0p?autoplay=true&startTime=1567053828000), so will file an issue/make a PR for it.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
BethanyG: Since the front end will be using React, I thought the constraint might be better there.  Also gives some flexibility if (and when!) the fields change....because they will...no matter how comprehensive they look now.  'Game' is one addition that comes to mind.  ;)

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
BethanyG: Agreed.  The classic way to model something like this would be with a many-to-many and a Tags table.  This was more a quick nod to the API specs, as well as an experiment with postgres JSONB capabilities.  So a bit of a hand-wave.

But as this system grows, I do have a few concerns/caveats with a "simple" many-to-many Tags table:

1.  Tags will most likely be a freeform field on the front end.  I don't really see our users choosing from an epic list of tags...and the "common convention" is for users to enter what they see fit.  This means writing a bunch of code to search for already existing tags and similar tags, and only add truly new/unique tags to the DB.  Even with that, the Tags table will potentially be a mess of overlapping and duplicate data from the get-go.  Just mocking up 28 resources for testing data created a "mess" -- and I am only one human.

2.  Learning Paths, Hangouts, Notes and Discussions will also be tagged, I would imagine.  So there will be potentially a lot of relations tracing back to the Tags table, each having overlapping, but not necessarily identical topics....so each would have to access the code for searching for existing tags/adding new tags.  With virtually everything connecting to the Tags table (and it being updated a lot per user, hangout, note, discussion, etc.), it could become a bottleneck fairly quickly.

3.  Postgres can index JSONB, and perform many of the same queries on JSONB that it can on regular data with small syntax changes.   Both autocomplete and recommendations can be done (albeit not at top speed) with documents.  But yes - if we are going the RDBMS route, then we shouldn't use JSONB too often, or in too many places.

One option is to use https://github.com/jazzband/django-taggit along with https://github.com/glemmaPaul/django-taggit-serializer, and customize the code so that each model has it's tags.  See this for a rundown and some thoughts:  https://medium.com/sthzg/a-short-exploration-of-django-taggit-bb869ea5051f.

And all that being said, I don't like how the Tags field adds an extra level right now, so I'm going to look at re-writing the serializer to omit that, and do some more digging on how Tags are implemented in other systems to see some more options.  Do you have a go-to model/resource you can share??  That would be great.

cbv3_django_prototype/cbv3_django_prototype/resources/views.py
BethanyG: Agreed.  Will add on my next revision.

cbv3_django_prototype/cbv3_django_prototype/resources/urls.py
BethanyG: Yup.  Will remove on next revision.

cbv3_django_prototype/config/settings/base.py
BethanyG: Oh, certainly.  This was done to avoid confusion because there weren't any instructions and many new users don't know how to deal with ENV Vars.  Would you mind putting some docs together on how you set/use Env vars for this?  I'll go ahead and change the settings file in the next iteration to use ENV Vars and make an issue for documentation.  Thanks!

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
brylie: Yeah, I think the backend should have the final say in validation. Adding choices constraint might also allow DRF to auto-document the allowed values.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
brylie: We ended up using a lot of JSONB where I work, and it has been a bit of a wrestling match. The main take-away, from a senior developer colleague of mine, is to only use JSONB for specific optimizations and to default to using a relational model with Postgres.

As mentioned, we can use an auto-complete widget along with the django-taggit package to improve consistency and avoid bespoke tagging code.

It might be worth checking out how Wagtail handles tagging, as it is a well-thought-out CMS:

https://docs.wagtail.io/en/v2.6.2/reference/pages/model_recipes.html
https://learnwagtail.com/tutorials/adding-tags-to-wagtail-cms-pages/

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
BethanyG: I guess I am not clear on how a CHOICES constraint works when we will not be using any front-end templates or views and will have all front end code in React.  But I will add on next update.

brylie: I believe the `choices` would be used to validate the input data:

> If choices are given, they’re enforced by [model validation](https://docs.djangoproject.com/en/2.2/ref/models/instances/#validating-objects)
>
> https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices

Also, we want to decouple the frontend from the server (as other clients may emerge with something like Vue, Flutter, etc), so the server should be able to stand alone -- including correctly validating data.

BethanyG: It appears to be a little more complicated than that...at least from my reading today.  But I am taking a look at it.  https://stackoverflow.com/questions/28945327/django-rest-framework-with-choicefield

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
BethanyG: I've re-written the tag field so that it doesn't have the weird extra level, but I am going to leave it as JASONB in the DB for now for several reasons:

1)  It is not clear in my reading & research that we'd take any sort of performance hit with using JASON in this fashion, and in fact several sources show JSONB to be preferable for speed and ease of use in a tagging model.

2)  This is currently a highly simplified model of what might be implemented in the DB should we go the Django/DRF route.  At this time, I don't necessarily want to complicate matters by adding 3 more tables to the mix, and JASONB gets our point across without them.

3)  I don't know that we have clarity around the importance/meaning/use of tags in our system, and I'd like to have some further discussions with others who are doing prototypes to hear how they envision it working.

4)  Changing the model a little later is not going to be terribly painful, since we are still in a prototyping phase.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
BethanyG: Here are the configured **RESOURCE_TYPES:**
        [('VID', 'Video'),
        ('POD', 'Podcast'),
        ('PODEP', 'Podcast Episode'),
        ('TALK', 'Talk'),
        ('TUTOR', 'Tutorial'),
        ('COURSE', 'Course'),
        ('BOOK', 'Book'),
        ('BLOG', 'Blog'),
        ('GAME', 'Game'),
        ('EVENT', 'Event'),
        ('TOOL', 'Tool'),
        ('LIB', 'Library'),
        ('WEB', 'Website')]

The validation will be on the first value, the second will be what is displayed in a UI or given back.  The Django Amin shows a drop-down for these values, but the DRF admin omits the field entirely.

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
BethanyG: Just so  that it is connected to this pull request, here is the JSON for the tags field:

"tags": [
                {
                    "id": "a1167632-abb8-4ce1-b090-b84e79c94c7f",
                    "name": "JavaScript"
                },
                {
                    "id": "66cc3539-6705-45ee-b15d-644d5fb90c41",
                    "name": "ES6"
                },
                {
                    "id": "8b99bcd9-a9a5-4455-b2da-fed06051d502",
                    "name": "React"
                }
            ]

cbv3_django_prototype/cbv3_django_prototype/resources/views.py
BethanyG: search_fields = ['id', 'media_type', 'title', 'description', 'tags']

cbv3_django_prototype/cbv3_django_prototype/resources/models.py
brylie: To expand a little bit, JSONB is good to used when we need 'data locality', in effect avoiding a second query or JOIN to get a related entity.

In general, Tags are used to show related content, e.g. "resources tagged Python". I believe that it will be desirable to use these tags to create views showing similar items, resources, etc.

The performance penalty comes when we try to search within the JSONB column (even with an index) for a tag string in the `name` property, and then join related documents together.

For these reasons, JSONB does not seem like a good approach for tagging, particularly when [the django taggit package makes a proper M2M tagging approach very straightforward](https://django-taggit.readthedocs.io/en/latest/getting_started.html).

Please excuse my directness, but I strongly recommend we use the django-taggit approach here or to omit the tagging feature from this pull request.

lpatmo: Let's leave tags implementation for a future issue -- I'll file @brylie's comments about using the `taggit` package (looks interesting!) and the performance penalty point makes sense.

> I don't know that we have clarity around the importance/meaning/use of tags in our system, and I'd like to have some further discussions with others who are doing prototypes to hear how they envision it working.

> Changing the model a little later is not going to be terribly painful, since we are still in a prototyping phase.

Agree -- it's not terrible if our model isn't perfect right now, as long as we don't forget to change it if we decide on continiuing with relational/postgres after the architectural comparison.

I think this PR is ready to merge in, but if there are other issues that should be filed (that I missed), please feel free to do so! Thanks again, all.

lpatmo: Also sorry @BethanyG, was scrolling back and realized you mentioned the `django-taggit` package first! That's what I get for skimming. :) Anyway, have filed the issue to follow up on tags when we need to here: https://github.com/codebuddies/django-concept/issues/16

PR #10, opened 2019-09-19
WPI: Initial setup of React in Django

lpatmo:
This is a WIP from the 40-minute stream at https://www.twitch.tv/videos/483344260

[ ] Is this the pattern we want to go with? An alternative is to separate the front-end entirely.
[ ] Make sure that this implementation works


brylie:
I think this should be in a separate repository, as it is still coupling the frontend code with the backend service.

brylie:
Basically, as long as all of the prototype backends adhere to the API specification, they can be developed independently of the backend code. This allows for experimenting with different client frameworks, like Vue or Flutter.

lpatmo:
Sounds good! This PR was based off of a particular Django+React pattern I found, but decoupling the front-end completely makes sense too.

PR #28, opened 2020-01-11
Remove virtual environment

### LINKED ISSUE #26: Clean up "cbv3" Python environment from repository
There is a folder in the repository root called "cbv3". It appears to be a local Python virtual environment. The virtual environment was likely checked in by accident, as it is unconventional to keep Python virtual environments under revision control.

## Task
- [x] Remove the "cbv3" Python environment from the repository

brylie:
Closes #26


brylie:
Ping @lpatmo @BethanyG

Issue #41, opened 2020-01-14
[Enhancement] Dockerize everything for dev env setup?


lpatmo:
## Context
A couple of people (@bengineerdavis, @d3vild06) have brought up/asked about docker-izing the entire app with docker-compose. My take was that I didn't think it was necessary, unless the benefits (easier setup) is very clear. @billglover explored it, and recorded a screencast of it here: https://codebuddies.slack.com/archives/C04BRN86J/p1578770421053800. @kennethlove offered some suggestions to tweak the approach.

## Acceptance Criteria
[ ] Explore a dockerized Django, Postgres, Mailhog, and Adminer (for viewing the DB). (Re: Mailhog -- that's fine! I expect at some point we'll be building out the transactional emails features)
[ ] Write a draft of instructions that users (like me) could test to set up the dev environment. Instructions should be friendly for those who've never used Docker before.

billglover:
Happy to pick this up.

Have removed the documentation label as this requires the addition of a `Dockerfile` and probably a `docker-compose.yaml`.

billglover:
For reference, this is the link to suggestions by @kennethlove: [Docker for Django](https://github.com/kennethlove/kennethlove.github.io/blob/master/content/posts/docker-for-django.md)

billglover:
Would be nice to re-name the app from `cbv3_django_prototype` to `app`. Any thoughts on this?

lpatmo:
Haha, that's an excellent idea. I'll file a Github ticket for it. The renaming (including the parts that need to be changed in the Dockerfile) could be all done in one go/one PR without any problems, right?

billglover:
Yeah, probably best to do this as a separate issue.

billglover:
Running into conflicting migrations with the latest merge. They aren't related to Docker, but I'll try to get this fixed so that we have a clean starting point.

lpatmo:
Oof, I remember Bethany saying she ran into issues with migrations too (near the end of the thread at https://codebuddies.slack.com/archives/C04BRN86J/p1579044879187600). I didn't because I had a clean install with no data and no migration files. 

Heading to sleep now, but let me know if you need help!

billglover:
I don't think it was that as I removed my database entirely before starting. Either way, migrations issue should be resolved and PR #44 should work. An edge case we may want to cover in documentation is that you'll need to ensure you don't have anything running that is already listening on these ports, e.g. an existing PostgreSQL install.

[...]

PR #44, opened 2020-01-14
Use Docker for local development environment

billglover:
Implement #41 by:

* [x] Providing `docker-compose up` as a way of instantiating a local development environment
* [x] Providing draft instructions for a new user to get an environment up and running

A couple of known issues:

* Static files are not served by the Django app. These should be moved to nginx.
* Migrations are run on start-up. We should provide a way to trigger these on demand.
* Tests are not run. We should provide a way to trigger these on demand.

I'm creating this pull request for discussion. I encourage developers to clone my fork and experiment with the development experience so that we can get some feedback before this is merged.


lpatmo:
🎉 🎉 
Instructions for how to fork this PR to test out docker:
- git remote add billglover https://github.com/billglover/django-concept.git 
- git fetch billglover
- git checkout billglover/master

lpatmo:
@billglover I tested it out by forking your fork into my personal account, and running `docker-compose up` looked really promising! :D Then I ran into this:

image

I believe it'll be fixed if you pull in the latest master changes -- e.g.`git pull upstream master` and `git push origin master` (assuming upstream points to the original CB django-concept). Do you mind trying that when you get a chance? There'll be a conflict in README.md (sorry about that); feel free to remove my version. 

(I tried that on my personal fork and re-ran `docker-compose up`, and got this error:
image --

which makes me think  we need `pip install -r requirements/local.txt` somewhere in docker, if we don't have it already.)

billglover:
I'll give it a go shortly. Dockerfile does include the `pip install` step. See: [Dockerfile#L10](https://github.com/billglover/django-concept/blob/master/cbv3_django_prototype/Dockerfile#L10).

I can see there is a conflict with the README, but I'll have a go at addressing this.

lpatmo:
Update: actually (in an attempt to make it easier) I submitted a PR to your fork with the latest as well: https://github.com/billglover/django-concept/pull/1

billglover:
I've merged the migrations, along with your PR. I've just run this on a clean machine and it seemed to stand-up CBv3 correctly. Would appreciate a couple of other people running through this and confirming it all works as expected.

d3vild06:
I am not sure if this is intended or not but I see a couple of things. 

1.  The folder structure seems odd where we have the same app folder nested inside with the same name:

Screen Shot 2020-01-16 at 11 56 45 AM


2.  The home page of the project after running `docker-compose up` is the following:

Screen Shot 2020-01-16 at 11 57 40 AM

Is this intended?

billglover:
I’m not sure either are desirable but neither are the result of using Docker for local development. Static content does need to be addressed in that we need to make a decision on how this is served.

lpatmo:
@d3vild06 A belated thanks for testing this out!

> The folder structure seems odd where we have the same app folder nested inside with the same name:
That's a [norm](https://django-project-skeleton.readthedocs.io/en/latest/structure.html) for Django, heh. One of the things that always bothered me. The project is called a certain name, and then we create an app inside it with the same name.

> The home page of the project after running docker-compose up is the following:
Ah, good pointing out -- I'll file an issue about adding some content to that Django home page to explain the relevant URLs after loading up the app -- namely `/api/v1` and `/admin` at this point, for example. We should make it clear that the Django project is mainly the API work that the React front-end will feed from.

billglover:
When running on Docker Desktop for Windows we have reports of the following error message:

```
ERROR: for db  Cannot create container for service db: C: drive is not shared. Please share it in Docker for Windows Settings
ERROR: Encountered errors while bringing up the project.
```

billglover:
The documentation here indicates that you need to specify which drives you want to use with Docker in the settings of the Docker App. This appears to be unique to Windows and is probably a result of the way virtualisation is implemented.

https://docs.docker.com/docker-for-windows/

To use volumes with Docker, users will need to explicitly share drives with the Docker App. For must users this will be the C drive but this is not guaranteed.

bengineerdavis:
@bill @lpatmo @d3vild06 Django can accommodate this change, if desired, and there's no penualty, should we want to move everything up a directory. Typically you can remove the extra nesting with (in a venv/container config file) `python django-admin startproject .` Adding the `.` is an optional argument for the django admin commands.

bengineerdavis:
@billglover, per Windows setup, maybe it should be explicit--if confirmed through tests--that setting a separate directory from the `C` or main directory for Docker is recommended to avoid system complications for Win10.

nuageklow:
Testing was performed on Windows 10 Pro with docker installed only based on the [setup instructions](https://github.com/billglover/django-concept).  

Also just ran `docker-compose up` in a venv with django installed (@bengineerdavis's suggestion) and `same error still came up cannot create container for service app: C: drive is not shared`.

billglover:
@nuageklow are you able to include a copy of the error you are seeing in your comment so that I can make sure we address it.

billglover:
I think I’ve found something that might make it easier for Windows users. If possible I need to be more explicit with referencing the current working directory. Will try and find a windows machine to test this on.

nuageklow:
@billglover 

`ERROR: for db  Cannot create container for service db: C: drive is not shared. Please share it in Docker for Windows Settings
ERROR: Encountered errors while bringing up the project.`

its the same error as before

lpatmo:
> Django can accommodate this change, if desired, and there's no penualty, should we want to move everything up a directory. 
@bengineerdavis Good to know, thanks! 

Re: Windows --  if all else fails, <s>Windows users can mainly contribute on the React app</s>  (JUST KIDDING! :P) I think we can make a note that Windows users should probably not use the `docker-compose approach`,  and follow a venv + local postgres approach.

nuageklow:
shared C: drive with docker on windows 10 Pro and an issue occurred when running `docker-compose up`:
```
app        |   File "/usr/local/lib/python3.7/site-packages/django/db/backends/postgresql/base.py", line 178, in get_new_connection
app        |     connection = Database.connect(**conn_params)
app        |   File "/usr/local/lib/python3.7/site-packages/psycopg2/__init__.py", line 126, in connect
app        |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
app        | django.db.utils.OperationalError: could not translate host name "db" to address: Name or service not known
```

sunu:
> shared C: drive with docker on windows 10 Pro and an issue occurred when running `docker-compose up`:
> 
> ```
> app        |   File "/usr/local/lib/python3.7/site-packages/django/db/backends/postgresql/base.py", line 178, in get_new_connection
> app        |     connection = Database.connect(**conn_params)
> app        |   File "/usr/local/lib/python3.7/site-packages/psycopg2/__init__.py", line 126, in connect
> app        |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
> app        | django.db.utils.OperationalError: could not translate host name "db" to address: Name or service not known
> ```
> 
> also noticed something that might be related to the issue above on creating db container:
> 
> ```
> db         | Data page checksums are disabled.
> db         |
> db         | fixing permissions on existing directory /var/lib/postgresql/data ... ok
> db         | creating subdirectories ... ok
> db         | selecting default max_connections ... 20
> db         | selecting default shared_buffers ... 400kB
> db         | selecting default timezone ... UTC
> db         | selecting dynamic shared memory implementation ... posix
> db         | creating configuration files ... ok
> db         | 2020-01-20 14:01:06.545 UTC [49] FATAL:  data directory "/var/lib/postgresql/data" has wrong ownership
> db         | 2020-01-20 14:01:06.545 UTC [49] HINT:  The server must be started by the user that owns the data directory.
> db         | child process exited with exit code 1
> db         | initdb: removing contents of data directory "/var/lib/postgresql/data"
> db         | running bootstrap script ... The files belonging to this database system will be owned by user "postgres".
> ```

I believe the first error is a caused by the second error. Since postgres failed to boot, db service can't be found. Using a named volume for postgres should fix the issue.

lpatmo:
@nuageklow @sunu Ah, from https://github.com/docker-library/postgres/issues/435, it looks like this is an example of having a named volume mount:

image

I forgot my husband has a Windows gaming computer, so tried it out there -- and all was smooth until I realized there are different editions of Windows, and his is a Windows _Home_ edition. For future reference, installing docker on Windows requires the following: `Windows 10 64-bit: Pro, Enterprise, or Education (Build 15063 or later).` And `Windows Home` users will need the following: https://docs.docker.com/toolbox/toolbox_install_windows/

If the named volume mount workaround doesn't work, I'm half-tempted to advocate the following: 
- https://docs.djangoproject.com/en/3.0/howto/windows/
- https://www.postgresql.org/download/windows/
... and suggest `docker-compose up` for other machines. 

Thoughts? Or am I giving up too early?

nuageklow:
> I forgot my husband has a Windows gaming computer, so tried it out there -- and all was smooth until I realized there are different editions of Windows, and his is a Windows _Home_ edition. For future reference, installing docker on Windows requires the following: `Windows 10 64-bit: Pro, Enterprise, or Education (Build 15063 or later).` And `Windows Home` users will need the following: https://docs.docker.com/toolbox/toolbox_install_windows/

@lpatmo just out of curiosity and claraification, was your husband successfully installed docker on windows 10 home edition and encountered the same issue on volumes when he was running `docker-compose up`?

> If the named volume mount workaround doesn't work, I'm half-tempted to advocate the following:
> 
> * https://docs.djangoproject.com/en/3.0/howto/windows/
> * https://www.postgresql.org/download/windows/
>   ... and suggest `docker-compose up` for other machines.
> 
> Thoughts? Or am I giving up too early?

I am probably not much of a help but im hoping to help on getting docker running on windows successfully (although i was trying on ubuntu yesterday and everything seems to be a lot smoother...). if its giving everyone trouble to move on maybe you guys can move on first and i'll try to figure this out on the side...will everyone know IF something comes up.

lpatmo:
> was your husband successfully installed docker on windows 10 home edition 
Sorry, to clarify -- I didn't get to that step of installing docker on Windows 10 Home because I only found out later that https://docs.docker.com/toolbox/toolbox_install_windows/ was a thing 😅

bengineerdavis:
I want to expand on some of my notes: in this scenario, file permissions can be changed either by Docker itself or by the application running inside of it. When I first started using Docker with Linux, I had database issues because Docker itself runs as `sudo` (superuser or admin for Windows), so any file that Docker accesses or creates will then only be allowed access at the root level, which would be a higher privilege than we normally have as a user after login. This is unique to Linux because Docker runs natively as a daemon process. This shouldn't be an issue for Docker on Mac/Linux, because Docker is made to run in a specialized container that is pre-configured for non-privileged access, meaning none of the processes ever run with root access.

That leaves us with the application itself running in the container. Postgres will only install with root privileges. This is part of how Postgres secures itself from potential hackers that get in through a regular user account. Because Postgres is changing files that would belong to a shared Docker volume. Since Postgres only runs in root (within Docker), then any file it access and changes will become root, too. That means that the shared volume is privileged even though our command docker-compose up and all the Django files are not. That's why you're getting issues with file access, I believe. I can speak to this because this gave me immense amount of headaches when I first worked with Django in Docker.

nuageklow:
yes I believe windows 10 pro / enterprise / education has something called Hyper-V which docker is required (maybe something Bill has similar issue with...but im not sure).

bengineerdavis:
https://docs.docker.com/compose/django/#create-a-django-project Has a note about a change in the Dockerfile for Django we need to add: `sudo chown -R $USER:$USER .`

bengineerdavis:
@nuageklow , can you please run a check on the files related to the docker project and confirm the file permissions and post an image of it here?

bengineerdavis:
Hyper-V is standard virtualization technology provided through the Windows platform.

bengineerdavis:
We won't need `venv` for this, because Docker is in itself isolating python dependencies at runtime. Basically, because Docker containers run in relative isolation from the rest of the operating system--except for when defined explicitly, such as mounted volumes--the `pip install -r requirements/local.txt` packages will already be isolated from our native install.

nuageklow:
> @nuageklow , can you please run a check on the files related to the docker project and confirm the file permissions and post an image of it here?

image
@bengineerdavis 
is this what you're looking for?

bengineerdavis:
@nuageklow I think we need to see permissions for the files in the repo you're using to bring up the app with `docker-compose`, but we also need to find out what the user permissions are for the volume being assigned to the Postgres container. I'm not sure how to specify that just yet (not extremely helpful, I realize).

I would try partitioning another drive on your computer, and then assign to to Docker for Windows, see what the result is. Depending, then we can go about working on file permissions.

nuageklow:
@bengineerdavis cb is the repo that contains docker-compose.yaml. 
image
let me know if there's a different way i can show you the permission of the repo file.

nuageklow:
i've figured out the solution!

the root of the issue for windows env is directory of the volume like everyone is indicating. apparently all we need to do is to change the the volume command from `./db_data:/var/lib...` to `db_data:/var/lib...` in the yaml file.

also windows requires a longer (detailed) command to call out a program, for example if i wanna pull psql i'll need to type:
`docker exec -it (container name) -d (database) -U (userid) -W` e.g. `docker exec -it db -d codebuddies -U babyyoda -W`

oh also since `./` is removed C drive doesnt need to be shared to install stuff within docker and it still worked properly.

can anyone please tell me how to run python manage.py shell with docker...? i would like to see if i can run the django app without issue.

bengineerdavis:
@billglover confirmed to workon Fedora 30 (Linux) OS with your setup.

Some error messages and adjustments observed:

- I first ran `docker-compose up` and got these error messages:
`WARNING: Image for service app was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Pulling nginx (nginx:1.17-alpine)...
1.17-alpine: Pulling from library/nginx
89d9c30c1d48: Already exists
534c802c70da: Pulling fs layer
ERROR: error pulling image configuration: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/36/36189e6707f43e39466f047ff9564fa55e217821aa7e2326b3f3c32166ca63b0/data?verify=1579585987-mf2TJMMidmlsTuxIEmAY4TkyK%2BI%3D: dial tcp: lookup production.cloudflare.docker.com on 192.168.1.1:53: read udp 192.168.1.7:47407->192.168.1.1:53: i/o timeout`

bengineerdavis:
Did not mean to close this issue!

lpatmo:
> can anyone please tell me how to run python manage.py shell (or any other regular django command) with docker...? i would like to see if i can run the django app without issue.

@nuageklow Can you try `docker-compose run --rm manage shell`?

sunu:
> Some error messages and adjustments observed:
> 
> 
> 1. I first ran `docker-compose up` and got these error messages:
> 
> > WARNING: Image for service app was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
> > Pulling nginx (nginx:1.17-alpine)...
> > 1.17-alpine: Pulling from library/nginx
> > 89d9c30c1d48: Already exists
> > 534c802c70da: Pulling fs layer
> > ERROR: error pulling image configuration: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/36/36189e6707f43e39466f047ff9564fa55e217821aa7e2326b3f3c32166ca63b0/data?verify=1579585987-mf2TJMMidmlsTuxIEmAY4TkyK%2BI%3D: dial tcp: lookup production.cloudflare.docker.com on 192.168.1.1:53: read udp 192.168.1.7:47407->192.168.1.1:53: i/o timeout
> 
This looks like a network timeout on docker registry's side. It should fix itself.
> 1. After I ran `docker-compose build` and then `up`, application ran. I then came across an error message for the `manage` service container:
> 
> > django-concept_manage_1 exited with code 0

This is kind of expected because the manage container is not exactly a log running service. If this is somewhat confusing, then we can write a Makefile command to start only the services needed. We can also create Make commands for common operations like running migrations or getting into the django shell so that people don't have to remember the long docker-compose commands. eg: `make shell` instead of `docker-compose run --rm manage shell`

docker-compose.yaml
```diff
@@ -0,0 +1,92 @@
+version: "3"
+
+volumes:
+  db_data: {}
+  nginx: {}
+  cbv3_django_prototype: {}
```
sunu: @billglover Just wanted to point out that we have all these named volumes defined but we're not really using these. Changing `./db_data` to just `db_data` on line 19, `./nginx` to `nginx` on line 55 and so on will make sure we use named volumnes instead of bind mounts. This should fix our problems on Windows too.

lpatmo: @sunu Are you saying that if @nuageklow makes those changes in the `docker-compose.yaml` file locally, it would probably work on her machine?

nuageklow: @lpatmo i actually posted the same solution lastnight earlier on this same thread..i guess i worded it in a different way..? 
https://github.com/codebuddies/django-concept/pull/44#issuecomment-576696357

but yes it worked and everything was installed properly.

for `docker-compose run --rm manage shell` command it didnt work. i'll find a way and post it if i have a solution.

sunu: @lpatmo Yes, that's my assumption. It seems they have already figured out as much in https://github.com/codebuddies/django-concept/pull/44#issuecomment-576696357

lpatmo: Oh sorry, I completely missed your comment from yesterday! Too many posts 😅 

So the remaining items to do are:
1/ Update the docker file in this PR
2/ Figure out the shell command
3/ sunu's `makefile` idea in a separate PR (maybe)?

billglover:
Working on an update today.

`Makefile` (I'm a big fan) should be a separate PR as we should discuss how Windows users will get the same (or equivalent) experience.

docker-compose.yaml
billglover: Slight nuance... making the change and removing the `./` to use volume mounts means volumes cannot be mounted from within the project folder and so code doesn't get mapped through to the running application. This means development on the host isn't reflected in the running application so won't work for us.

I'm looking through documentation now to see if we have to drop back to bind mounts to achieve the set-up we are working towards.

billglover:
This post goes some way to explaining the behaviour we are seeing here: [Define named volume with host mount in the docker compose file](http://blog.code4hire.com/2018/06/define-named-volume-with-host-mount-in-the-docker-compose-file/).

Unfortunately, I think we need to make bind mounts work with Windows.

docker-compose.yaml
nuageklow: @sunu @billglover do you think it would be something windows specific only? i was sort of worried abt things might not be placed properly even thought it seemed like it ran smoothly. 

also any expected end result be included would be ideal (e.g. how to define the volume is installed properly) sorry i guess im too new on this..

cbv3_django_prototype/requirements/local.txt
```diff
@@ -27,3 +27,4 @@ django-debug-toolbar==2.0  # https://github.com/jazzband/django-debug-toolbar
 django-extensions==2.2.1  # https://github.com/django-extensions/django-extensions
 django-coverage-plugin==1.6.0  # https://github.com/nedbat/django_coverage_plugin
 pytest-django==3.5.1  # https://github.com/pytest-dev/pytest-django
+django-taggit==1.2.0 # https://github.com/jazzband/django-taggit
```
sebbel: what is this package used for?

sebbel:
so props to @billglover. I could instantly contribute to the repository with the help of this docker setup. kudos!

billglover:
Confirmed working on Windows 10 Pro. I have been able to complete the following:

* start the application
* access the admin interface in the browser
* make changes to the template and see them reflected in the browser
* log in to Adminer and view database schema
* log in to the Mailhog interface
* run Django `manage.py` commands

I propose we run the fix for #53 once this is merged.

If anyone would like a walk through of this running on Windows or Mac OS, please message me directly on Slack.

cbv3_django_prototype/requirements/local.txt
lpatmo: @sebbel `django-taggit` is a tag management library that makes creating a global tagging system easier. Docs here: https://django-taggit.readthedocs.io/en/latest/

lpatmo:
Thank you so much for your work on this, @billglover! 🙏 And also to everyone who helped test this out! How do you feel about merging this, and we could file separate issues for:
1/ Windows Home users 
2/ creating a makefile

Then:
1/ #53 - clean up .gitignore
2/ Makefile to shorten commands
3/ https://github.com/codebuddies/django-concept/pull/54 can be updated

billglover:
I think this is ready to merge.

I would suggest leaving Windows 10 Home until we have a need for it, but a Makefile and/or script for Windows users would be well worth it. Agree with 1, 2, 3.

sunu:
Love how well-documented everything is @billglover!

sunu:
The only remaining nitpick I have is to change restart policy of the containers from `always` to `on-failure`. `always` automatically starts the services even after shutdown and reboot which is not really desirable on local machines; especially laptops.

billglover:
@sunu This isn’t a nitpick. This is a really good shout. I pulled this together on a desktop that I predominately access remotely. Completely overlooked the fact that the majority will be working on laptops. Will get that changed.

PR #64, opened 2020-02-14
Add Postman collection

### LINKED ISSUE #62: Use a shared Postman Collection
I've noticed a couple of contributors using Postman to test calls to the APIs exposed as part of the CodeBuddies backend. I would like to explore the idea of a shared Postman collection.

**Why?** By sharing a Postman (or equivalent) collection we should hope to see a couple of potential benefits. We would gain the ability to consistently call the API making it easier to reproduce observed behaviour. We should also expect to see new contributors finding it easier to make calls to the API.

**Why now?** Contributors have already started building up personal Postman collections. These get harder to merge as the collections grow.

**Cost of doing nothing?** If we don't adopt a shared collection we would leave it up to contributors to put together a method for making API requests. These could be documented as part of API documentation, or something as simple as a list of cURL commands.

billglover:
Fix #62 by creating a skeleton Postman collection.

The skeleton collection enables the following features:
* use random data for all requests
* create user (capture token, username and password in variables)
* login (capture token in variable)
* list resources (use token for auth)
* create resource (use token for auth)

### DIFF
--- a/postman/CodeBuddies.postman_collection.json
+++ b/postman/CodeBuddies.postman_collection.json
@@ -1,0 +1,233 @@
+{
+	"info": {
+		"_postman_id": "5b8c55c5-59e8-4e5b-a4c4-4fe5e0386dec",
+		"name": "CodeBuddies",
+		"description": "# CodeBuddies\n\nThis collection contains requests for CodeBuddies backend APIs.",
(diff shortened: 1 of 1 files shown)


billglover:
Ready for review.

lpatmo:
@sebbel, @albertfougy, and I reviewed in a hangout!

PR #89, opened 2020-02-24
Issues 30 31 file reorg part 2

brylie:
Here are a few additions to the PR by @lpatmo 

Note: this pull request is based on @lpatmo's branch, so changes in this PR will go into that branch :-)


lpatmo:
@brylie Thank you!!

Another issue, unfortunately: 

1/ I checked out this branch

2/ When I run `docker-compose up --build`, I see this error:

image

3/ I opened up my project files and saw that there seems to be an extra cbv3_django_prototype folder next to the `projects` folder you created:

image

docker-compose.yaml
```diff
@@ -26,14 +26,14 @@ services:
...
   # allow development to be done on the host machine.
   app:
-    build: ./cbv3_django_prototype
+    build: ./project
     container_name: app
     restart: on-failure
     command: >
-      sh -c "python /opt/cbv3_django_prototype/manage.py migrate &&
-             uwsgi --ini /opt/cbv3_django_prototype/uwsgi.ini"
+      sh -c "python /opt/project/manage.py migrate &&
```
billglover: This looks like it should reference `/opt/codebuddies/...`.

brylie:
> I opened up my project files and saw that there seems to be an extra cbv3_django_prototype folder next to the projects folder you created

@lpatmo try opening my branch in a clean clone of the repo. There may be some files remaining when switching branches.

docker-compose.yaml
brylie: Good catch, thanks!

lpatmo:
Will do!

Update: I deleted `backend`, did a git clone of the repo again, and then checked out the `issues-30-31-file-reorg-part-2` branch.

Ran `docker-compose up --build` and saw the following:

image

This is confusing because I know ./project/manage.py exists!

```
Lindas-MacBook-Pro-2:backend lpnotes$ find . -name "manage.py"
./project/manage.py
```

This is what I see in docker-compose.yaml when I have the codebase open:

image

Sorry for the back and forth; not sure what is wrong. :/

brylie:
The command should be located in `/opt/codebuddies/manage.py`. I think I updated the Dockerfile, but may have overlooked some places where it is using `/opt/project` instead of `/opt/codebuddies`. Make sure you are using the latest commits from this branch as well.

lpatmo:
IT WORKS!!! 

Scarily, the first time I tried -- I git pulled and `docker-compose down` and `docker-compose rm` and `docker compose up --build` -- was still seeing the errors.

So I blew away the folder and re-cloned the repo, but then checking out this branch and `docker-compose up --build` was fine. 

Only issue seems to be these missing `static` files, but they can be added easily:

image

Going to merge this and run this on the other branch, and hope a `git pull` will suffice vs. having to delete and re-clone the repo 🤞

lpatmo:
Thanks again for helping with this!

PR #91, opened 2020-02-28
Deploy to Staging and Production

### LINKED ISSUE #65: Deploy the master branch to production
We don't yet have a production deployment of the CodeBuddies backend. We should have one.

**Why?**:
Deploying a publicly accessible version of the CodeBuddies backend will allow us to get feedback from the CodeBuddies community.

**Why now?**:
The earlier we get feedback, the easier it is to address. We have functionality that is very close to being ready for user testing. It would be good to have an environment to deploy to in time for release.

**What if we don't implement?**:
User testing will require users to stand up their own environments. This makes it less likely that a user will use features to accomplish real tasks.

## Acceptance

* [x] publicly accessible deployment of the CB backend
* [x] ~publicly accessible deployment of the CB frontend~
* [x] tagged changes to `backend/master` trigger a deployment of the CB backend

*Edit:* Removed frontend from the acceptance criteria as that is being deployed differently.

billglover:
Fix #65 by creating two GitHub Actions Workflows; Stage and Deploy.


billglover:
You can see links to the deployed environments here:
https://github.com/codebuddies/backend/deployments

BethanyG:
This is next on my list..but I need  a nap first.  I will review this weekend, however.

billglover:
I'll update this with the changes in #96 once they are merged in.

billglover:
I've merged in the latest changes from master. @codebuddies/backend this is ready for review.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: start deployment
+        uses: bobheadxi/deployments@master
```
lpatmo: Interesting -- how did you find https://github.com/bobheadxi/deployments, out of curiosity?

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,72 @@
...
+        id: deployment
+        with:
+          step: start
+          token: ${{ secrets.GITHUB_TOKEN }}
```
lpatmo: Is this `secrets` object defined in Dockerhub or Digitalocean?

project/config/settings/production.py
```diff
@@ -63,53 +41,6 @@
     "DJANGO_SECURE_CONTENT_TYPE_NOSNIFF", default=True
 )
 
-# STORAGES
-# ------------------------------------------------------------------------------
-# https://django-storages.readthedocs.io/en/latest/#installation
-INSTALLED_APPS += ["storages"]  # noqa F405
```
lpatmo: ++ Removing cruft from production file! yay

.github/workflows/stage.yaml
```diff
@@ -0,0 +1,85 @@
...
+          desc: Setting up staging deployment for ${{ steps.build_image.outputs.tag }}
+
+      - name: copy files
+        uses: appleboy/scp-action@master
+        with:
+          host: ${{ secrets.DO_STAGING_HOST }}
+          key: ${{ secrets.DO_SSH_KEY }}
+          passphrase: ${{ secrets.DO_SSH_PASSPHRASE }}
+          port: ${{ secrets.DO_STAGING_SSH_PORT }}
+          username: ${{ secrets.DO_SSH_USER }}
```
lpatmo: Are you generally thinking that both `staging` and `production` would be hosted under one DO droplet, but have different DBs?

.github/workflows/deploy.yaml
billglover: I searched the Actions marketplace. I've no prior experience of this, but it seemed to do the trick. I picked this one as it seemed to focus on manipulating the deployments API rather than doing the deployment.

.github/workflows/deploy.yaml
billglover: These secrets are [defined in GitHub](https://github.com/codebuddies/backend/settings/secrets). You enter them once but can't see the values again. They get presented to your actions for use during execution.

project/config/settings/production.py
billglover: I was unsure about these as they clearly serve a purpose. I removed what I had to in order to deploy with the production configuration. It is entirely possible these will be used later. My working assumption is that we can add these back if needed.

.github/workflows/stage.yaml
billglover: For now, yes. They would run as two containers on the same small droplet. The reason I separated out the secrets is that I believe we may want to run on separate droplets at some point in the future.

billglover:
Questions are good as they force me to challenge what I was thinking when pulling this together. You are right though, testing is difficult without actually deploying this. I fully expect to see future issue look to tweak this, particularly around what triggers a deployment and how we handle data.

.github/workflows/deploy.yaml
lpatmo: Gotcha, that makes sense! I only found out a few days ago that Actions is so new that GitHub is sponsoring a three-week remote hackathon for it, btw -- https://githubhackathon.com/. :)

/doesthismeanwe'remodern

.github/workflows/deploy.yaml
lpatmo: Gotcha, that makes sense, thanks!

.github/workflows/deploy.yaml
billglover: They've been around in beta for a while, but they definitely got better with the roll-out.

billglover:
Removed the target droplet from DO to save cost. Am closing this PR for now as it has fallen behind master but can re-open when we want to tackle deployment to staging/prod.

PR #92, opened 2020-02-29
[Tagging 71-74-86]  Cleaning up Some Tagging Issues 

### LINKED ISSUE #71: [API] Non-English (unicode, extended ASCII, emoji) Characters Cannot Be Slug-ified in Tags
During the review of PR #69, @billglover uncovered a bug with `tags` that include Non-English characters (Unicode, extended ASCII, emoji):  https://github.com/codebuddies/backend/pull/69#discussion_r380910342

In addition to not respecting the characters, the resulting strings fed to the serializer screw up the ordering and return of the `tag:slug` pairs, and are saved into the DB in garbled form.

On the surface, this appeared to be a relatively straightforward problem:  have Django or taggit respect the non-english characters.  It is, however a rather complex situation. 

[**Original PR filed for taggit**](https://github.com/jazzband/django-taggit/issues/347)
[**Second PR filed for taggit**](https://github.com/jazzband/django-taggit/pull/315)
[**linked issue (still open)**](https://github.com/wagtail/wagtail/issues/3311)
[**linked proposal for dealing with the issue (still open)**](https://github.com/wagtail/wagtail/pull/5508)  
[**additional open issue blocking the above**](https://github.com/wagtail/wagtail/pull/3088)

___

Since we cannot rely on `django-taggit` to come up with an "easy" solution to this (_nor do we have the forms and front end code complications that team does_), we'll have to find a way to apprehend the strings and translate/transliterate or otherwise format them in a way that they can be slugified appropriately.  

This may mean using an alternative `slugify` process, or altering/patching the `taggit` code.  [Here](https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py) is a link to the `taggit` code that saves tags to the DB, as a place to start.

Not super-clear on the path forward.  First thoughts are 

1.  Apprehend the strings ahead of `taggit`, and translate or transliterate them (not optimal unless we can re-translate them for appropriate display)

2.  Alter `taggit` code to slugify the strings in a way that can be stored in the DB and retrieved in their original unicode form, since we do not have the restriction of making them nice for URLs, etc.

### LINKED ISSUE #74: [API]  Tags is a Required Field for POST-ing a Resource
Currently, the model for `Resouces` has `tags = TaggableManager(blank=True)`, and while this will allow a *blank* `tags` field, it will not allow a poster to **omit** the `tags` field from the POST.

This is not desirable, and we will need to consider if we would like to write code to apprehend the incoming POST to check for the missing field and supply it , as opposed to throwing a "this field is required" error...or otherwise research how to make the code more flexible so that the field is not required.

### LINKED ISSUE #78: Faker package installation error
Run the following command to install development packages:

```plain
pip3 install -r /opt/cbv3_django_prototype/requirements/local.txt
```

During package installation the following error is returned:

```plain
ERROR: faker 4.0.1 has requirement text-unidecode==1.3, but you'll have text-unidecode 1.2 which is incompatible.
```

Expected behaviour: no errors or warning shown during package installation.

### LINKED ISSUE #86: [API]  Store Tag Serializer Code as a **tagging** app for Use Across Endpoints
Currently, we are assuming most endpoints will have **Tags** and we are also using `taggit` for tag management.  

Our current `taggit` model required custom serializer code to make tags for `resources` display appropriately, and the expectation is that this same code (or some improved or different version) will be required for any endpoint that is using `taggit` or tags.

We should therefore move the **TagSerializer** and **TagsSerializerField** code out of `resources`, and into a common ~~utility file or~~ location for ease of import and use to other endpoints (and to better discuss and maintain it as a module).

~~Propose it be moved to the **`utils.py`** file, where we also have the jwt response handler code...or another common or core place.~~

Propose a **tagging** app for storing tagging related serializers, views, and functions.

BethanyG:
- Moves tagging related functionality to a new **`tagging`** app.  (fixes #86) 
- Fixes bugs with unicode rendering of tag slugs (fixes #71) 
- Adds a UUID1 **`guid`** for tags.  (fixes #71 )
- Removes the requirement for a **`tags`** field when POSTing a resource.  (fixes #74 )
- Bump **`python-slugify`** version to 4 (fixes #78)


BethanyG:
Yeah.  I think this PR needs to be hard-booted, and I need to start over with less dubious history.  10-1, the repo size is due to the "recovery turdlets" git left on my system after I had to run **`git fsck`.  

The migrations are a longer story, and one I'd welcome help on.  I am going backup and do a hard clear of my local repo, and a refresh of my fork and re-apply these individual changes.  I'll try to get a new (smaller!!) PR for testing before EOD.

BethanyG:
I am going to close this as unsalvageable, and will open a new PR when things are reset.

PR #93, opened 2020-03-01
Taggingredux 71 74 86

BethanyG:
_Once more, with feeling_

### Procedure for re-doing this:

1.  Refreshed fork from `master`, made sure `master` overwrote the fork
2.  Deleted local repo
3.  Cloned fork to local machine
4.  Branched and re-did the code edits

*Note:*  In order to get the migration order to work correctly, adding dependancies to the migration files  was necessary per documentation [**here**](https://docs.djangoproject.com/en/3.0/howto/writing-migrations/#controlling-the-order-of-migrations)  and adding the **`__latest__`** flag to dependent migrations as noted [**here**](https://stackoverflow.com/questions/29416956/how-to-get-the-last-migration)

-------

- Moves tagging related functionality to a new **`tagging`** app. (fixes [#86](https://github.com/codebuddies/backend/issues/86))
- Fixes bugs with unicode rendering of tag slugs (fixes [#71](https://github.com/codebuddies/backend/issues/71))
- Adds a UUID1 **`guid`** for tags. (fixes [#71](https://github.com/codebuddies/backend/issues/71) )
- Removes the requirement for a **`tags`** field when POSTing a resource. (fixes [#74](https://github.com/codebuddies/backend/issues/74) )
- Bump **`python-slugify`** version to 4 (fixes [#78](https://github.com/codebuddies/backend/issues/78))
- Adds dependancies on **`__latest__`** `taggit` and `tagging` migrations to **`resources`** migration file.


BethanyG:
*sigh* So the trick that worked for me locally is not working in docker.  Evidently, the migration files for `taggit` are named differently when loading into docker, so the file dependancies I put in to make things work.....don't.  

I think I need  a break from this atm, but essentially, we want migrations to run in this order:

1.  taggit
2.  tagging
3.  resources
4.  everything else

**HOWEVER** if you do not put dependancies in the migration files for Django to use, and you run those migrations manually, it changes the **`contenttype_id`** of resources ... and then the tagging fixtures and tests break...just an added special bonus of fun.  I will take another look after a break for a nap and some dinner.

BethanyG:
Putting an additional note here:  should you need to take a look at the migration files that **`taggit`** is keeping, they are (well mine are) located here `../lib/python3.7/site-packages/taggit/migrations`.  In other words, in the `site-packages` sub-directory.

billglover:
I can confirm that this does fix: #78. This also no longer has the large repo size issue that we saw with #92.

I'm hitting migration issues with the following error:

```plain
django.db.migrations.exceptions.NodeNotFoundError: Migration resources.0006_auto_20200229_1516 dependencies reference nonexistent parent node ('taggit', '0004_taggeditems')
```

Interestingly I don't have a `../lib/python3.7/site-packages/taggit/migrations` folder. Is this a strange case of a dependency on something that lives outside the repository? I would expect all tag migrations to sit in `./tagging/migrations/` as this is where tags are defined.

cbv3_django_prototype/cbv3_django_prototype/tagging/admin.py
```diff
@@ -0,0 +1,17 @@
+from django.contrib import admin
+from resources.models import TaggedItems
```
billglover: I'm surprised to see a dependency on `resources.models` inside the `tagging` app. I would expect `resources` to depend on `tagging` and not the other way around. Interestingly this appears to be used on line 8 so I'm not sure I can just remove the dependency.

lpatmo:
> I'm hitting migration issues with the following error:
> 
> ```
> django.db.migrations.exceptions.NodeNotFoundError: Migration resources.0006_auto_20200229_1516 dependencies reference nonexistent parent node ('taggit', '0004_taggeditems')
> ```

(Can confirm the same error.)

Thanks for this. Will review more tomorrow!

cbv3_django_prototype/cbv3_django_prototype/tagging/migrations/0001_initial.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import migrations, models
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('taggit', '0003_taggeditem_add_unique_index'),
```
lpatmo: Elementary question: what does this `0003` number represent?

lpatmo: (It's auto-generated by Django, but I wonder why I'm seeing `0004_taggeditems` elsewhere. Maybe this is not an important detail.

billglover: The numbers are given so that migrations can be executed in the order they were developed. They are auto incrementing and this is why I would never expect to see modifications to migrations.

cbv3_django_prototype/cbv3_django_prototype/tagging/admin.py
BethanyG: This is case where Dango's file structure can be misleading.  While this file is parked inside the **`tagging`** app **directory**, its code is instructions to the **`admin`** app on how to display tags & tagging in the admin interface correctly.  I think these get placed with their apps so that there's no confusion on which piece of the admin they affect..but that creates confusion of a different type.

So while **`resources`** needs **`tagging`** to tag,  and **`tagging`** needs classes from **`taggit`** to override,  **`admin`** needs all three to be able to pull and display **`resources`** with their related **`tags`**:

admin interface

cbv3_django_prototype/cbv3_django_prototype/tagging/migrations/0001_initial.py
BethanyG: The numbers are auto-assigned by Django as the ORM generates the changes needed for the database.  They're generated/added to each time **`makemigrations`** is run **for the system you are on**.  

image

...migrations are not run in the order they are developed -- the dependency tree is resolved **at runtime** by the Django ORM.  This usually works fine in the case of straightforward additive changes...and breaks entirely with things like moving a manytomany relationship around without specifying an ordering, or needed to override how a third party app is behaving (like our situation).  

image

Please see [**controlling the order of Migrations**](https://docs.djangoproject.com/en/3.0/howto/writing-migrations/#controlling-the-order-of-migrations) for more info.

The big problem here is that these auto-generated and auto-numbered files are done when **`makemigrations`** is run .... and we don't have any control over how Django does that for third party Django apps, so we don't have control over what numbers get assigned to **taggit**s migration files on install.

I am going to see if just specifying the **0001** file forces the correct order.  We know, at least that **0001** will always be generated.

BethanyG: _**I spoke too soon**_  See [**getting the last migration**](https://stackoverflow.com/questions/29416956/how-to-get-the-last-migration) for more info.  Will push a fix shortly.

BethanyG: Apologies.  To also answer the earlier question:
 **'0003_taggeditem_add_unique_index'** is the third migration file for **`taggit`**.  

The **`0004_taggeditem`** references are for the 4th migration file for the **tagging** app, which inherits from, and is dependent on, **taggit**.

BethanyG:
ok.  Testing is still broken which tells me that the ordering is still not quite working.  Bear with me...I am doing one more push.

BethanyG:
So I am having trouble understanding why the docker build process won't respect the order of migrations I've specified in my migration files.  All of the errors I am seeing during the docker build process are (again) related to the order in which the migrations are being run.  What am I missing here??

billglover:
Are there migrations we need to run that aren’t stored in the Git repo? This might be a source of confusion.

The Docker build process doesn’t do migrations. On start-up we run:

```plain
python /opt/cbv3_django_prototype/manage.py migrate
```

Is this not what gets run locally?

BethanyG:
It is what gets run locally, but it appears that it is running in an entirely different order than when I do this on my local system.  Here is the last thing I tried:

1.  Switched to new branch that is a clone of `taggingredux 71 74 86`
2.  Deleted all migration files from **resources** and **tagging**
3.  Ran **`docker-compose up -d --build`**.  Get warning about migration files missing, but otherwise everything works.
4.  Log into the running **app** container.  Run **`python manage.py makemigrations`**
5.  Run **`python manage.py migrate`**  - things explode
6.  Run each migration separate by calling **`manage.py migrate taggit`** , followed by **tagging** and then ** resources.   This succeeds.
7.  Stop everything.  Go to local conda env.  
8.  Drop local db.  Create local db.
9.  Go to conda env.  Run **`manage.py migrate`** with files generated from running the **`makemigrations`**  command from within docker
10.  Everything succeeds.

So..I have to run errands and do some other stuff.  I won't be able to look at this any more until later.   I don't know how to make docker respect the order.

BethanyG:
Sorry - the answer is YES there are migrations that are run that are not in the git repo.  A ton of them.  Django runs migrations for all of its framwork apps - none of which are really present in the repo.  This includes contenttypes, taggit, admin...and I cannot remember what else.

billglover:
Ouch... it is going to be very difficult to manage database changes if these exist outside the repository.

This seems really counter intuitive and doesn't follow the documentation here: [Migrations](https://docs.djangoproject.com/en/3.0/topics/migrations/).

From poking around it seems as if these should live in the `./contrib/*/migrations/...` folders but we only have migrations for a `sites` app inside our repo.

billglover:
In the case of `taggit` migrations seem to be generated in the following location: `/usr/local/lib/python3.7/site-packages/taggit/migrations`.

billglover:
OK so these are the migrations run when deploying using Docker:

```plain
account
 [X] 0001_initial
 [X] 0002_email_max_length
admin
 [X] 0001_initial
 [X] 0002_logentry_remove_auto_add
 [X] 0003_logentry_add_action_flag_choices
auth
 [X] 0001_initial
 [X] 0002_alter_permission_name_max_length
 [X] 0003_alter_user_email_max_length
 [X] 0004_alter_user_username_opts
 [X] 0005_alter_user_last_login_null
 [X] 0006_require_contenttypes_0002
 [X] 0007_alter_validators_add_error_messages
 [X] 0008_alter_user_username_max_length
 [X] 0009_alter_user_last_name_max_length
 [X] 0010_alter_group_name_max_length
 [X] 0011_update_proxy_permissions
contenttypes
 [X] 0001_initial
 [X] 0002_remove_content_type_name
django_celery_beat
 [X] 0001_initial
 [X] 0002_auto_20161118_0346
 [X] 0003_auto_20161209_0049
 [X] 0004_auto_20170221_0000
 [X] 0005_add_solarschedule_events_choices
 [X] 0006_auto_20180322_0932
 [X] 0007_auto_20180521_0826
 [X] 0008_auto_20180914_1922
 [X] 0006_auto_20180210_1226
 [X] 0006_periodictask_priority
 [X] 0009_periodictask_headers
 [X] 0010_auto_20190429_0326
 [X] 0011_auto_20190508_0153
resources
 [X] 0001_initial
 [X] 0002_auto_20200116_1610
 [X] 0003_auto_20200116_1619
 [X] 0004_auto_20200118_1014
 [X] 0005_auto_20200118_1036
 [ ] 0006_auto_20200229_1516
sessions
 [X] 0001_initial
sites
 [X] 0001_initial
 [X] 0002_alter_domain_unique
 [X] 0003_set_site_domain_and_name
socialaccount
 [X] 0001_initial
 [X] 0002_token_max_lengths
 [X] 0003_extra_data_default_dict
tagging
 [X] 0001_initial
taggit
 [X] 0001_initial
 [X] 0002_auto_20150616_2121
 [X] 0003_taggeditem_add_unique_index
userauth
 (no migrations)
users
 [X] 0001_initial
```

The reason `0006_auto_20200229_1516` fails appears to be this error:

```plain
  Applying resources.0006_auto_20200229_1516...Traceback (most recent call last):

...

  File "/usr/local/lib/python3.7/site-packages/django/db/backends/base/schema.py", line 524, in alter_field
    not new_field.remote_field.through._meta.auto_created):
AttributeError: 'str' object has no attribute '_meta'
```

I can't find the `str` object anywhere.

BethanyG:
Apologues - I am still running errands.    The error that resources is throwing relates to the name of the through table that is listed in the arguments passed to the tags field in the resources model.

Django can’t resolve it because it is running resources 0006 before it has created the referenced table, which is created during the migrations for tagging and taggit.  I *think* the name it is looking for is under  taggit 0003.


I’ll be home in an hour and can look closer then.

-B

BethanyG:
image


That "meta" messages is being thrown because the Resources model has the following line:

image


..which means that field is dependent on both the the `CustomTaggableManager` (from **tagging**) and the `TaggedItems` (from both **tagging** AND and alter on **taggit**)...but if the migration for `resources` is being run before those other two can get things sorted....then `resources` can't find the class names and app names associated with the DB tables.  Or at least that's how I interpret that.

Feeling pretty sick, so I think I need to take a pause and look at this with fresh eyes in the morning - but I am seeing quite a few SO posts around migrations getting pretty wonky when they're tipped off in docker, so that's where I am looking next.  Specifically, this issue:  https://stackoverflow.com/questions/33992867/how-do-you-perform-django-database-migrations-when-using-docker-compose.  And since I've noticed that the HTML interfaces aren't working, I am wanting to try `collectstatic` anyways.

BethanyG:
I'm going to rethink this.

Issue #48, opened 2020-01-15
[Testing]Rewrite Application Tests to Use Factory-Boy


BethanyG:
The current Application tests load Fixture data for JWT and Resource testing.  These should now be re-written to use Factory Boy (or Model Mommy), so that they are not dependent on static or moldy  data.

BethanyG:
Curious as to why this is an enhancement?  These tests are broken in current code.

billglover:
Would be great if these tests could:

* Be executed with `manage.py test`
* Populate whatever data they need to run
* Remove that data once complete

BethanyG:
**NOTE** Model_mommy is no longer maintained.  [**Model Bakery**](https://pypi.org/project/model-bakery/) is the recommended migration.


AFIK, `manage.py test` via `unittest` and `PyTest` takes care of:

* Creating a test DB, 
*  Loading (or creating, depending on the libraries used) data into the test DB, 
*  Running the tests (provided they are prefaced with **test** in their name)

..but weather or not the loaded data is generated on the fly or is static  (**fixtures**) depends on how the tests are written.  Right now, they are super basic and pretty fragile.

billglover:
> "Model Bakery offers you a smart way to create fixtures for testing in Django. With a simple and powerful API you can create many objects with a single line of code."

BethanyG:
Since I do not have strong opinions for using one of these over the other, I'm going to go with [**Factory-Boy**](https://factoryboy.readthedocs.io/en/latest/), both because @chris48s is familiar with it, and because it works for projects beyond Django, and is therefore a good testing tool for potential contributors to learn.   

Open to additional opinions or discussion, but changes beyond this point need to include Chris and any others who are writing tests so that their work isn't  invalidated or broken by a library decision.

BethanyG:
@chris48s  -- we also have [**Faker**](https://faker.readthedocs.io/en/master/) installed, but I don't really like what I am reading about how limited the data it produces is.  Granted, I haven't done _extensive_ reading.  Any recommendations there are very welcome.  😄

billglover:
Thank you. This feels like we have a decision to progress with Factory-Boy. Do we want to keep the issue open until we have updated existing tests?

BethanyG:
@billglover  -- yes, I think that's our current strategy.

PR #96, opened 2020-03-04
Taggingfixes 71 74 86 ..and 78

BethanyG:
*Note:*  In order to get the migration order to work correctly, adding dependancies to the migration files  was necessary per documentation [**here**](https://docs.djangoproject.com/en/3.0/howto/writing-migrations/#controlling-the-order-of-migrations)  and adding the **`__latest__`** flag to dependent migrations as noted [**here**](https://stackoverflow.com/questions/29416956/how-to-get-the-last-migration)

-------

- Moves tagging related functionality to a new **`tagging`** app. ( fixes [#86](https://github.com/codebuddies/backend/issues/86))
- Fixes bugs with unicode rendering of tag slugs ( fixes [#71](https://github.com/codebuddies/backend/issues/71))
- Adds a UUID1 **`guid`** for tags. (fixes [#71](https://github.com/codebuddies/backend/issues/71) )
- Removes the requirement for a **`tags`** field when POSTing a resource. ( fixes [#74](https://github.com/codebuddies/backend/issues/74) )
- Bump **`python-slugify`** version to 4 (fixes [#78](https://github.com/codebuddies/backend/issues/78))
- Adds dependancies on **`__latest__`** `tagging` migrations to **`resources` 0006_auto_20200303-1257.py** migration file.  
-  Adds **`tagging.json`** & **`taggeditems.json`** files to populate the custom_tags and taggeditems tables.
-  Fixes broken **`resources`** tests.


lpatmo:
> adding the __latest__ flag to dependent migrations  
> https://stackoverflow.com/questions/29416956/how-to-get-the-last-migration

TIL 

Thank you again for the detailed notes!

BethanyG:
@billglover - adding you for review to make sure I haven't screwed up something in my battles with the ORM & docker.  Ok if you don't want to give specific feedback..just not quite ready to merge this without additional eyes.

billglover:
Will try and get to this in the next 48 hours. Have been tied up.

cbv3_django_prototype/cbv3_django_prototype/resources/tests.py
```diff
@@ -9,10 +9,10 @@
 class ResourcesTests(APITestCase):
 
     def setUp(self):
-
         call_command('loaddata', 'users.json', verbosity=0)
         call_command('loaddata', 'resources.json', verbosity=0)
-        call_command('loaddata', 'tags.json', verbosity=0)
+        call_command('loaddata', 'tagging.json', verbosity=0)
+        call_command('loaddata', 'taggeditems.json', verbosity=0)
```
kennethlove: Instead of loading fixtures, I'd recommend either using `pytest` fixtures, a library like `factory_boy` to generate data, or just creating the needed model instances here in the `setUp`. Fixtures get out of date quickly and are fairly hidden. It's also hard to know if you can change them to create new data conditions for tests. This would also solve your problem of needing to fetch items below as you could associate the created objects with the class and then use them in tests below.

BethanyG: Thank you for the reminder.  We have #48 out as a specific issue around this...the logic being that I was struggling to get tagging related features done (and wanted at least some very minimal tests to go with them), so bailed on the testing infrastructure in the short term.

I will be adding additional issues for testing models, serializers, and views as well now that the "major" changes are done-ish.

BethanyG:
oop.  Brain fart on the merge process.  Mashing big, green button now....

PR #100, opened 2020-03-07
Avoid starting mailhog and adminer during testing

### LINKED ISSUE #68: [Testing]Improve Test Workflow execution time
The [Test Workflow](https://github.com/codebuddies/backend/actions?query=workflow%3ATest) currently starts two optional applications during testing. Neither Mailhog or Adminer are required. Starting them during the workflow consumes unnecessary resources and may slow down execution.

Acceptance:
* [ ] Only required services started during Test Workflow: `web`, `app`, `db`
* [ ] Testing continues to execute and results remain visible

billglover:
Fixes #68

- Starting unnecessary services takes time and consumes resources
- Remove the need to start optional services; `mailhog` and `adminer`
- Users may need to add the `--remove-orphans` flag the first time they
  run `docker-compose down` after this change. Docker will alert users
  if this is the case.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,23 +8,23 @@ on:
 
 jobs:
   test:
-
     name: test
     runs-on: ubuntu-latest
-          
+
     steps:
-    - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
 
-    - name: Run application
-      run: docker-compose up -d
-    
-    - name: Test Resources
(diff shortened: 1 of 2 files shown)


billglover:
> Am I still meant to see adminer and mailhog start up?

You are when you run this locally yes. This change removes them during tests that run on commit.

lpatmo:
> This change removes them during tests that run on commit.

Ahh, I totally missed that. Got it, thanks!

PR #102, opened 2020-03-08
Unblock cors error on react app running on localhost:3000

lpatmo:
If I have the django and react app running at the same time on localhost:8000 and localhost:3000 respectively, I get a cors error if I try to make requests to localhost:8000/auth/users from the React app:

image

This PR whitelists http://localhost:3000 to prevent the cors error.


lpatmo:
> How do we guarantee that users run the React app on port 3000

Heh, good question. The default is 3000, but folks can definitely specify alternate ports. I'll make sure to document on the React side. Thanks for reviewing!

Issue #84, opened 2020-02-22
[API]  Register Projects Endpoint with Admin Interface and Create Admin Views


BethanyG:
We will want the ability to review/administer the **Projects** endpoint through the Django Admin interface, and so will want to register our `projects` app with the Django admin and provide any custom admin views we think we'll need.

See [**The Django Admin Site**](https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#overview), [**Admin Actions**](https://docs.djangoproject.com/en/3.0/ref/contrib/admin/actions/), and [**Customizing the Admin**](https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#customizing-the-adminsite-class) for more details on how to start.

See [**Resources : admin.py**](https://github.com/codebuddies/backend/blob/master/cbv3_django_prototype/cbv3_django_prototype/resources/admin.py) for an example of how the `resources` endpoint customized the Admin to display tags appropriately. 

Since `projects` will also include a potential **manytomany** relation with **users**, there may need to be additional customization around that for the Admin interface.

BethanyG:
As of **PR** https://github.com/codebuddies/backend/pull/96, tags serialization is handled by an app, **`tagging`**.  Below is an example from **`resources/admin.py`** on how **tagging** is used:

```python
class ResourceAdmin(admin.ModelAdmin):

    list_display = ['tag_list']

    def get_queryset(self, request):
        return super().get_queryset(request).prefetch_related('tags')

    def tag_list(self, obj):
        return u", ".join(o.name for o in obj.tags.all())

admin.site.register(Resource)
```

BethanyG:
Closing as no longer relevant to current development.

watchtheblur:
working on it now

watchtheblur:
Django admin errors out with "Internal Server Error" when I have the following code in admin.py
```python
class OSProjectAdmin(admin.ModelAdmin):

    list_display = ['tag_list']

    def get_queryset(self, request):
        return super().get_queryset(request).prefetch_related('tags')

    def tag_list(self, obj):
        return u", ".join(o.name for o in obj.tags.all())
```

BethanyG:
@watchtheblur - I think you're missing a couple of pieces of info in the osprojects/admin.py file.

First, (and I always forget this!) you need to import your model at the top of the file so that Django knows what DB tables are going to be administered through the admin interface.  For `osprojects`, the two imports should be:

```python
from django.contrib import admin
from .models import OSProjects
[...]

PR #105, opened 2020-03-10
serve static assets from nginx

chris48s:
At the moment, the backend throws a 404 on any javascript or CSS files.
This means the admin interface looks like this:

Screenshot at 2020-03-10 20-34-33

There is clearly intent to use make use of the admin backend e.g: https://github.com/codebuddies/backend/issues/84 . Enabling nginx to serve static assets makes it much more usable:

Screenshot at 2020-03-10 20-33-58

..which is nicer :)

In general, this is primarily an API project so the vast majority of production traffic will be serving JSON. Javascript and CSS is only likely to be served to a small handful of admin users so I'd propose this as a production setup. There's not really any point in spending time on a more elaborate solution like farming it out to a static object storage service (S3/Spaces/whatever) to optimise the 1% case.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -30,7 +30,8 @@ services:
     container_name: app
     restart: on-failure
     command: >
-      sh -c "python /opt/codebuddies/manage.py migrate &&
+      sh -c "python /opt/codebuddies/manage.py collectstatic --clear --no-input &&
+             python /opt/codebuddies/manage.py migrate &&
              uwsgi --ini /opt/codebuddies/uwsgi.ini"
     volumes:
       - ./project:/opt/codebuddies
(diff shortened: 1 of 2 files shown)


BethanyG:
Nice!  Thank you so much for doing this!  We would have eventually gotten to it..er..maybe.  
But you fixed these, and now the Admin and "browsable API" are things of beauty.  YAY!:

image


image


Agree that since this is probably the limit of "stuff" we'd be serving up as web pages, we don't really need a more elaborate setup for storage ATM.

BethanyG:
@bill - tagging you, just in case there's something I've missed or other setup/config considerations you might have.

lpatmo:
YAY thank you @chris48s!! 🚀

billglover:
Yey! 🚢 

Let's get these merged. I'll take care of updating open PR for staging deployment separately as I need to figure out to deploy these.

BethanyG:
Mashing the big, green button.  🚢 🚀

Issue #70, opened 2020-02-19
Document the use of the makemigrations command


lpatmo:
## Summary
I want to run `pip install -r requirements/local.txt` while inside the `backend_app` docker container so that I can install the `taggit_serializer` module, but I can't get inside it because my containers are continuously restarting.

## Recreation of steps

$ docker-compose down
$ docker-compose rm
$ docker-compose up (with no -d)

Result:
![image](https://user-images.githubusercontent.com/4512699/74854701-88de7680-52f4-11ea-8696-3e77dd06125b.png)

From those error messages, I saw that a module `taggit_serializer` was not installed.

In a non-docker environment, normally I would `pip install -r requirements/local.txt` (or `pip install -r requirements.txt assuming that defaults to running the local.txt file which is specific to cookiecutter) while I am inside a virtualenv.

But because I am running docker, I'll have to run that command inside a container. 

From https://stackoverflow.com/questions/56904783/how-to-install-a-python-library-to-a-pulled-docker-image, I found that I could run:

$ docker ps #check for the container ID
$ docker exec -it container_id bash

So I tried this:
![image](https://user-images.githubusercontent.com/4512699/74855094-1326da80-52f5-11ea-8a3f-939bdadc6b3f.png)

But I keep getting the message that the container is restarting on the `backend_manage` and `backend_app` containers. 

![image](https://user-images.githubusercontent.com/4512699/74855144-246fe700-52f5-11ea-90ee-3777da6dc2e4.png)

Am looking at https://stackoverflow.com/questions/37471929/docker-container-keeps-on-restarting-again-on-again next for hints on what I can do to solve the "container is always restarting" problem 😅

billglover:
Will add more detail later...

```plain
docker-compose build
```

I think the following may also work but need to check.

```plain
docker-compose up --build
```

The trick is knowing when this needs to be run. That’s where I get stuck. We could say run it all the time, we could say run it only if certain things change, or we could create a `Makefile` (Windows users may miss out here).

lpatmo:
Ahhh thank you so much, `docker-compose up --build` worked!! :D I noticed it also applied migrations; do you know if it also does `makemigrations`?

> The trick is knowing when this needs to be run.
Generally, we need to `pip install -r requirements/local.txt` whenever a new Python module is added. And we need to `makemigrations` and `migrate` each time the DB model changes. 

So maybe the tip would be to run it after each `git pull` from the master branch, in case there have been model changes or new modules added? (hand-wavy)

[...]

PR #106, opened 2020-03-10
fail the build if migrations are not up-to-date

chris48s:
refs #70

This change will cause the build to fail if there changes to a model which have no corresponding migration. Its still important to document it, but this automates the check for it in CI.


BethanyG:
Did the following to test this:

1.  Pulled down the branch
2. Made a change to the resources model and pulled it into a test branch
3. Checked in and committed the change without running **`makemigrations`**
4.  Opened a PR as a test
5.  Got the following during the build:

image

So LGTM!

PR #109, opened 2020-03-12
Run the entire test suite using pytest

### LINKED ISSUE #107: [Testing] docker-compose run --rm manage test users runs zero tests (PyTest not used as TestRunner)
The `users` app does have some tests defined in it: https://github.com/codebuddies/backend/tree/master/project/users/tests but they're not actually being run in CI: https://github.com/codebuddies/backend/blob/1607b816ad9c36c2fc0428c081b4d8bf914b25da/.github/workflows/test.yml#L24-L26

```
$ docker-compose run --rm manage test users
Starting db ... done
System check identified no issues (0 silenced).

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
```

The main problem here is that there is a mix of pytest tests and django tests. Django's `unittest.discover` (which is what is being used in CI) isn't picking up the pytest ones.

In general, the project should standardize on one test runner for discovery and that is best done sooner rather than later. I'm happy to help out solving this one way or the other, but is anyone able to give a quick summary of how you ended up here and where you're trying to get to so I know which direction to go in?

chris48s:
Main changes in this PR:

:wrench:  Configure pytest
:eyes:  Get it to discover and run the full test suite
:heavy_check_mark:  Get all the existing unit tests passing
:books:  Update CI build and docs

Closes #107


project/config/settings/test.py
```diff
@@ -27,11 +25,6 @@
     }
 }
 
-# PASSWORDS
-# ------------------------------------------------------------------------------
-# https://docs.djangoproject.com/en/dev/ref/settings/#password-hashers
-PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
```
chris48s: The reason I've done this is because the passwords in the fixture files were encrypted using the default hasher, so this was causing all the auth-based tests to fail (the hashes didn't match). This didn't manifest when running the tests with django's own runner because it wasn't set up to use this file at all - it was using `local.py`.

Another approach would have been to rebuild all the fixtures, but this was easier (and makes a minimal difference to test suite speed). If you end up switching all your tests to create test data on-the-fly using `FactoryBoy` or whatever, you could re-enable this micro-optimisation at a later date.

chris48s:
Hrmm. This is close, but not quite ready for review. I've marked this WIP for now as I've run out of time to fiddle with it this evening and there's clearly a bit of a hole in my docker knowledge that needs plugging to finish this off and tidy up. I'll have to come back to it in a few days now.

Remaining tasks are:

- Clean up the commands for invoking `manage.py` and `pytest` calls in the CI build
- Reflect that solution in the docs
- Rebase to clean up the history

@billglover - I think the issue I'm hitting is that essentially `docker-compose run --rm manage [x]` is running as a less privileged user than `docker-compose run --rm app [x]`. I just can't explain why, or tbh what is the logic behind `manage` as a separate service from `app`. Are you able to offer any insight on this that might help?

chris48s:
Having fiddled with it a bit more, I think this is good for review now. I've ended up removing the `manage` container. I did have a look at turning it into a generic `shell` container for running shell commands, but it ended up being an exact clone of the `app` container so I decided we might as well have one less service and run stuff against the app container.

I think my original comment in https://github.com/codebuddies/backend/pull/109#issuecomment-598461820 was wrong - everything is running as root. Simultaneously, I feel like I've worked around an underlying issue here which stems from the containers running as root and writing to the filesystem but I think in its current form this PR isn't making that any worse than when I started. I'd still like to understand it better though :thinking:

BethanyG:
Hi @chris48s - I am going to dig into this (and some testing) today.  I don't think I feel comfortable removing the `manage` container without more eyes on this, so I've requested that @billglover take a look as well.

README.md
```diff
@@ -89,17 +89,22 @@ If you would like to tail the logs in the console then you remove the detach fla
 
 The following are examples of some common Django management commands that you may need to run.
 
-* Make Migrations: `docker-compose run --rm manage makemigrations`
-* Merge Migrations: `docker-compose run --rm manage makemigrations --merge`
-* Run Migrations: `docker-compose run --rm manage`
-* Test: `docker-compose run --rm manage test`
+* Make Migrations: `docker-compose run --rm app ./manage.py makemigrations`
+* Merge Migrations: `docker-compose run --rm app ./manage.py makemigrations --merge`
```
lpatmo: What does `--merge` do? Is this the right documentation? https://docs.docker.com/engine/reference/commandline/app_merge/

README.md
```diff
@@ -89,17 +89,22 @@ If you would like to tail the logs in the console then you remove the detach fla
...
+* Run Migrations: `docker-compose run --rm app ./manage.py`
```
lpatmo: I think this is supposed to be `docker-compose run --rm app ./manage.py migrate` (missing the "migrate" :))

project/config/settings/base.py
```diff
@@ -257,7 +257,10 @@
             "level": "DEBUG",
             "class": "logging.StreamHandler",
             "formatter": "verbose",
-        }
+        },
+        "null": {
```
lpatmo: Can you explain what this does? (Sorry if it is obvious and I missed it)

project/users/models.py
```diff
@@ -10,6 +10,5 @@ class User(AbstractUser):
     # around the globe.
     name = CharField(_("Name of User"), blank=True, max_length=255)
 
-    @property
```
lpatmo: Why did we need to remove this decorator, out of curiosity?

README.md
chris48s: I just cargo culted it from the existing docs which were added in https://github.com/codebuddies/backend/commit/f85bc3b81960542c32152a0124daf6b18cf8281b
The right docs are for that flag are: https://docs.djangoproject.com/en/3.0/ref/django-admin/#cmdoption-makemigrations-merge
Its a flag to do with squashing and merging migrations, but I'm not 100% sure why that specific flag is documented in this README.

README.md
chris48s: :+1: Good call fixed in 98b567a

project/config/settings/base.py
chris48s: In general, f6eddc6 is declares a logger that throws away log messages and then says "use this logger for everything under test"

I originally put that there when I was running tests with the django runner to capture log messages under test because they were all getting printed to stdout, but didn't clock that once I switched them over to running with pytest, pytest does this for us in a more configurable way. I've backed this change out in cb766dd. We should rely on pytest's capturing: https://docs.pytest.org/en/latest/capture.html
`pytest` will capture the logs by default and `pytest --capture=no` (or `pytest -s`) won't, which is more useful anyway.

I've done cb766dd as a rever commit to make review easier, but I can tidy up history before merge

project/users/models.py
chris48s: This is explained in the commit message for 03770f4

To work through a more concrete example of the problem this causes:

property:
* Declare `get_absolute_url()` as a property
* Go to http://127.0.0.1:8000/admin/users/user/
* Click on any user
* Click the "view on site" link in the top-right
* It will throw `'str' object is not callable`: you can't invoke a string.

method:
* Now switch it to a method
* Click the "view on site" link in the top-right - it will issue a redirect.


Side note: It doesn't issue a redirect to a _useful_ URL because of https://github.com/codebuddies/backend/blob/2683d9a3d9fb3e51287712ae5ce184c178242206/project/core/contrib/sites/migrations/0003_set_site_domain_and_name.py#L16 but lets deal with that as another issue. I'm trying to avoid scope-creep too much in this PR and just focus on the tests. The reason I fixed the signature in this PR which addresses testing (rather than treating it as a seperate bug) is because it was causing a test to fail. Its just that nobody has noticed because you weren't running all the tests.

README.md
lpatmo: Ah, thank you for finding the docs for that `--merge` command! 🤔

README.md
lpatmo: Thanks!

project/users/models.py
lpatmo: Thanks for the explanation here.

BethanyG: On closer inspection, we should remove the redirect method altogether, & also remove the `[view on site] ` button from the Admin.  (see this for how:  https://stackoverflow.com/questions/8662359/django-remove-view-on-site-button-in-the-admin-user-change-form). I think this decorator + method was either left from Coookiecutter or cargo-culted in by myself when piecing together userauth.  Since we are mostly building an API, the redirect for viewing user detail/user profile on the site won't  _ever_ work -- unless we point it at a react URL.  If we do provide a user profile or user detail, it will be as an endpoint, and deliver JSON -- not an HTML template/form.

project/config/settings/test.py
BethanyG: We should have a community discussion about which hashing method we want to use for passwords in the DB - we're using **Argon2** ATM (recommended in the Django docs - but not the default, which is PBKDF2), which is why the fixutres have the hashes they do.  So I recommend using `django.contrib.auth.hashers.Argon2PasswordHasher` so that the two match.

project/users/models.py
BethanyG: I'll open an issue for that (and probably chuck it into the PR for our testing re-write)...unless there are any objections??

BethanyG:
@billglover - any thoughts or concerns to add?

billglover:
No concerns with removing the `manage` container. The tests all seem to pass and so I'm happy with this.

Not much to add regarding `pytest` or the changes to configuration and application tests as I'm still a long way behind on Python / Django familiarity.

project/users/models.py
chris48s: The 'view on site' button is probably not that helpful for an API project, but I picked it as an example of an internal call that expects `get_absolute_url` to be a method which exists at the moment, but I think there are other internal calls in django (and possibly DRF) that expect that signature.

If you reeeally want a property version, we could also declare

```py
@property
def absolute_url(self):
    return self.get_absolute_url()
```

> and probably chuck it into the PR for our testing re-write

It can be done as its own PR - should be no dependency/merge conflict there

chris48s:
@billglover - do you have any thoughts on the issue of files created inside the container being owned by root? Should we be chown-ing everything from outside the container, should the containers be running as not-root? Bit of both?

billglover:
I’m not too fussed at the moment, but would like to revisit how the containers are built and see if we can reduce the number of things that need to be installed. Perhaps this is would be a good time to look at file ownership.

Issue #115, opened 2020-03-23
[Testing] [WIP] Use Factory-Boy to Create  Seed Data for Dev and Staging DBs


BethanyG:
This needs more detail, but I am logging this here so I don't forget it.  Will update with detail shortly.

chris48s:
Now #116 is merged I'm happy to pick this one up. I think its pretty straightforward:

:factory:   Write a management command to build some test data using factories
:x:  Delete the fixture files
:books:  Update the docs with how to init DB with test data

Are you happy for me to just run with it, or do you want to pair on this one too?

BethanyG:
Hey!  Now I have emoji bullet envy.  😄 

Every time I think of this issue, I get hung up on introducing some sort of "meaning" to the data...like I want to make it "pretty" for display, or have a set of "realistic" resources with related tags...or resources that would have different groupings for search....and that is probably not a good thing.  I think my brain is just stuck on the fixtures that I use out of habit.  🙄   

So YES YES - more than happy to have you take this up.  Have fun playing with it.....and since you are doing it programmatically, it would be great to make a default batch of say 50+ or so ...or maybe even a configurable batch?

We may also want to do something like scrape tech related urls to select from..or not.  How about I back away slowly and let you get to work?  Yup, that's what I should do.  😉

PR #116, opened 2020-03-25
Rewrite application Tests using Factory-Boy

### LINKED ISSUE #48: [Testing]Rewrite Application Tests to Use Factory-Boy

chris48s:
Closes #48

Done so far:
- [x] Rewritten the tests in the `userauth` app to use the `UserFactory` (the `user` app tests generated by cookiecutter already had a `UserFactory` class) instead of fixtures
- [x] Written a `CustomTagFactory` class and sketched out what the `TaggedItemsFactory` needs to look like

The remaining tasks are:
- [x] Write a `ResourceFactory` class and finish off `TaggedItemsFactory`
- [x] Rewrite the tests for the `resources` app to use the tag/resource factories instead of fixtures.

We might also look at generating init data (#115) in this PR, or we could leave it for another PR.

@BethanyG - I'll mark up the diff with some notes explaining what I've done so far a bit more. I'll leave it with you to have a look at this PR, work on the `ResourceFactory` and `resources` tests and push some more commits to this branch.


project/userauth/tests.py
```diff
@@ -10,18 +12,11 @@
...
     def setUp(self):
-        # create a new user
-        model = get_user_model()
-        self.person = model.objects.create_user(
-            username='PetuniaPig',
-            email='pretty.piglet@pigfarm.org',
-            password='codebuddies'
+        self.user = UserFactory(
+            password=PostGenerationMethodCall('set_password', 'codebuddies')
         )
```
chris48s: The reason I'm doing this rather than

```py
self.user = UserFactory(password='codebuddies')
```

is because we need the value to be hashed.

project/tagging/factories.py
```diff
@@ -0,0 +1,34 @@
+import uuid
+from django.contrib.contenttypes.models import ContentType
+from factory import DjangoModelFactory, Faker, LazyAttribute
+from .models import CustomTag, TaggedItems
+
+
+class CustomTagFactory(DjangoModelFactory):
+    guid = LazyAttribute(lambda obj: uuid.uuid1())
+    name = Faker("word")
```
chris48s: Lets keep this simple and deal with:
* Setting up a load of example data in our init script
* Explicitly testing cases like "हिंदी में जानकारी" and "🐸"

in the tagging tests, as opposed to relying on loading fixtures in the resource tests to do it indirectly.

project/tagging/factories.py
```diff
@@ -0,0 +1,34 @@
...
+https://factoryboy.readthedocs.io/en/latest/recipes.html#django-models-with-genericforeignkeys
+our TaggedItemsFactory is probably going to look roughly like this:
+
+from resources.factories import ResourceFactory
+
+class TaggedItemsFactory(DjangoModelFactory):
+
+    content_type = factory.LazyAttribute(
+        lambda o: ContentType.objects.get_for_model(o.content_object))
+    content_object = factory.SubFactory(ResourceFactory)
```
chris48s: When we create a factory with a default value of `SubFactory`, the default is that creating a `TaggedItems` with `TaggedItemsFactory()` will also create an associated `Resource` object, but we could also invoke `TaggedItemsFactory(content_object=foo)` where foo is an existing `Resource` object if we need more control over the associated resource.

project/tagging/factories.py
chris48s: We don't need to fake slug here - we'll create one on `save()` based on the name

mayela:
Hi @chris48s and @BethanyG. I'm Mar. I added the ResourceFactory, hope you can review it. Thanks in advance

project/resources/factories.py
```diff
@@ -1 +1,40 @@
...
+
+
+class ResourceFactory(DjangoModelFactory):
+    guid = LazyAttribute(lambda obj: uuid.uuid1())
+    title = Faker("sentence")
+    author = Faker("name")
+    description = Faker("text")
+    url = Faker("uri")
+    referring_url = Faker("uri")
+    other_rerring_source = Faker("uri")
```
chris48s: typo here

```suggestion
    other_referring_source = Faker("uri")
```

project/resources/factories.py
```diff
@@ -1 +1,40 @@
...
+    ('TUTOR', 'Tutorial'),
+    ('COURSE', 'Course'),
+    ('BOOK', 'Book'),
+    ('BLOG', 'Blog'),
+    ('GAME', 'Game'),
+    ('EVENT', 'Event'),
+    ('TOOL', 'Tool'),
+    ('LIB', 'Library'),
+    ('WEB', 'Website')
+]
```
chris48s: There are a couple of issues here:
First is we want to pass `choice()` a straight list, not a list of tuples (unlike django's `choices` which accepts a list of tuples).
Second is we don't really want to declare and maintain this list in two places. If we change it in the model we want our factory to update too

Lets use a list comprehension and declare `RESOURCE_TYPES` as:

```py
RESOURCE_TYPES = [
    r[0] for r in Resource.RESOURCE_TYPES
]
```

That would would solve both of those problems.

project/resources/factories.py
```diff
@@ -1 +1,40 @@
...
+
+
+class ResourceFactory(DjangoModelFactory):
+    guid = LazyAttribute(lambda obj: uuid.uuid1())
+    title = Faker("sentence")
+    author = Faker("name")
+    description = Faker("text")
+    url = Faker("uri")
```
chris48s: :+1:

project/resources/factories.py
BethanyG: @chris48s - excellent suggestion.  I only have one quibble here (and yes, I am being a cranky-pants dev....BUT) single-letter variable names make me _squirm_.  They almost never help readability..especially in a comprehension.  Would be much happier with: 

```python
RESOURCE_TYPES = [resource[0] for resource in Resource.RESOURCE_TYPES] 
```

...or something along those lines.  😉

mayela:
Thanks for your comments @chris48s. Today I'm going to review each one and resolve it. Nice day! :blush:

chris48s:
There's one remaining task to finish off this PR, which is to rewrite the resources tests to use the `ResourceFactory` added by @mayela . Are you still planning to pick this up @BethanyG ?

BethanyG:
@chris48s  - I am.  I have just been snowed under.  Should be able to get to it before the weekend.

chris48s:
No pressure :)

mayela:
@BethanyG I want to help with tests! How we can work together?

mayela:
@chris48s Sorry, I didn't notice I closed the PR! :sweat:

BethanyG:
@mayela  - We could always use more tests and more testing coverage!!  I'll open an issue, and we can go from there....  also -- you can always open issues here in the repo for things you think we need to put in place, and if it's work you want to work on..just assign the issue to yourself.  We can always discuss details/use cases as we go. 😄 

I went ahead and re-wrote the existing (and sorta pathetic) `resources` tests so we can get this PR closed...but we can (and need to) add more tests and scenarios around user auth, resources, tagging, and of course, integration tests.  You can also touch base with @chris48s and see if there are things on his mind you could help out with as well.

project/tagging/factories.py
BethanyG: Yup.  Hard agree here.  Much better to test tagging from _**tagging**_ and only briefly touch on it in related endpoints.  (Methinks that we do need to test that the "connection" is made in a trivial way - that I can add **a** tag at the least to a `resource`, `hangout`, or `learning path` via POST or PATCH..but not more than that.).  Everything else should be in the tagging tests.

BethanyG:
@lpatmo  - adding you here as a reviewer because I've just done a push..so need additional eyes on this.  @chris48s  - I cannot seem to add you as a reviewer, or I would have.  😉 .

lpatmo:
Thanks for tagging me; I will take a look!

project/resources/tests.py
```diff
@@ -1,88 +1,110 @@
...
         data= {
             "description": "A _diabolically irresponsible_ talk in which I celebrate modern Python coding by **abandoning all backwards compatibility** THE END",
-            "tags": ["test tags", "testing"]
+            "author": "David Beazley"
         }
+
         response = self.client.patch(url, data, format='json')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data['description'], data['description'])
-        self.assertEqual(response.data['tags'][0]["name"], data['tags'][0])
```
chris48s, line 58: What happened to the assertions about tags here? (not necessarily wrong to bin them - its just not explained anywhere, I don't think)

project/resources/tests.py
```diff
@@ -1,88 +1,110 @@
...
+        self.assertEqual(response.data['author'], data['author'])
 
 
     def test_delete_one_resource(self):
-        #to do:  grab the guid from the created guids in the DB instead of hard coding it.
-        url = '/api/v1/resources/96eec6e2-59f4-11ea-9149-dca9047779fe/'
+        records = create_batch(ResourceFactory, 10)
+        new_resource = create(ResourceFactory, guid='96f4dd16-59f4-11ea-9149-dca9047779fe')
+        url = f'/api/v1/resources/{new_resource.guid}/'
```
chris48s, line 79: Because you're using `new_resource.guid`, we don't actually need to hard-code the guid. This could just be:

```py
new_resource = create(ResourceFactory)
url = f'/api/v1/resources/{new_resource.guid}/'
```

Also, same comment for `test_view_one_resource()`

chris48s:
> I cannot seem to add you as a reviewer,

I think its because I opened the PR: I can't approve my own PR (even though 3 of us have now pushed to the branch)

project/resources/tests.py
BethanyG: Yup.  That's due to #125 😄 .  I didn't even think to do a skip -- I'll add that in, I like how you did it with the issue ref in [49eca65](https://github.com/codebuddies/backend/commit/49eca650bbc4b1ec152201a7783bb5f32c349cf5) .

BethanyG: as for binning the tags..I was solving an issue with the ordering being different coming back in `data` .. and then decided I didn't want to.  But I should do at least one test case with tags, so I can add it back in.

project/resources/tests.py
BethanyG: Good point. Dunno what I was thinking there...I did that elsewhere, but brain farted here.  Fixed in the new push.  Thank for catching!

BethanyG:
@chris48s - do you wanna have one last look and then push the big, green button?  _**Thank you so much for your patience and work on this!**_. 🌟

chris48s:
LGTM :+1:
We can move on to #115 now :)

Issue #98, opened 2020-03-04
[Testing] Add Model Tests for Resources and Tagging


BethanyG:
Need to add specific model tests for Resources and Tagging, and provide framework going forward for those types of tests for other endpoints.

- [x] CustomTag model - https://github.com/codebuddies/backend/pull/121
- [ ] Resource model

Adding these links, in case they're helpful:

-  [Testing Models with Django using PyTest and Factory-Boy](https://medium.com/@dwernychukjosh/testing-models-with-django-using-pytest-and-factory-boy-a2985adce7b3#:~:text=Pytest%20and%20Factory%20Boy%20make,running%20testing%20logic%20on%20them.&text=The%20django_db%20decorator%20must%20be,to%20access%20the%20testing%20models.)
-  [Real Python:  Django PyTest Fixtures](https://realpython.com/django-pytest-fixtures/)
-  [Django Stars: Testing Your App with PyTest](https://djangostars.com/blog/django-pytest-testing/) -- longer and more detailed article on PyTest in General and tips for Django testing with PyTest
-  [PyTest Docs](https://docs.pytest.org/en/stable/contents.html)
[pytest-django Docs](https://pytest-django.readthedocs.io/en/latest/)

BethanyG:
still open.

Issue #72, opened 2020-02-19
[Testing] Need to Add Specific Tests for Tagging Scenarios


BethanyG:
Currently, tags are being tested as a side effect of `Resources`.  Given the issues we've already uncovered, we need to create specific tests for tagging and tag formatting that are pulled out into their own tag test files.  We also need test data for these tests.

Not an exhaustive list, but here are some use cases we need to test for:

1.  Non-English extended ASCII characters as part of tags (see bug #71 )
2.  Unicode characters as part of tags
3.  "Special" characters (slashes, semicolons, anything that might read as code) as part of tags
4.  Nested JSON in tags
5.  Nested lists in tags
6.  Malformed or otherwise weirdly formatted tags
7.   Duplicate tags (to test de-duping)
8.  Various capitalization scenarios
9.  Adding tags to an existing set
10. Removing tags from an existing set

BethanyG:
Yes - the tags on the issue are weird.  But it is both a bug we don't have tests, and and enhancement that we add all the scenarios.

lpatmo:
👏 Great list. There's validation the front-end form can do of course, but would be good to validate if we're making API requests directly too. I have no idea how to implement testing against duplicate tags in DRF so this would be good learning for me...

BethanyG:
Pulling this into a branch where I am working on multiple proposed changed to tags.

PR #121, opened 2020-03-29
add tests for tag slugging

chris48s:
refs #98
refs https://github.com/codebuddies/backend/pull/116#discussion_r398159174

build will be failing for now.. see comments


project/tagging/tests.py
```diff
@@ -1,3 +1,44 @@
...
+            {"name": "自動化する", "expected_slug": "自動化する"},
+            {"name": "sjálfvirkan", "expected_slug": "sjálfvirkan"},
+            {"name": "پژوهش ", "expected_slug": "پژوهش"},
+            {"name": " గ్రాఫ్", "expected_slug": "గరఫ"},
+            {"name": "데이터 베이스", "expected_slug": "데이터-베이스"},
+
+            {"name": "हिंदी में जानकारी", "expected_slug": "-जनकर"},
+            {"name": "प्रयास है", "expected_slug": "रयस-ह"},
+            {"name": "stòran-dàta", "expected_slug": "sstòran-dàta"},
+            {"name": "స్వయంచాలకం", "expected_slug": "యచలక"},
```
chris48s: In these four cases, I've copied the names and slugs from the fixture: https://github.com/codebuddies/backend/blob/master/project/tagging/fixtures/tagging.json but the tests fail. @bethanyg - can you confirm what the expected slugs are in these 4 cases? Have we spotted a bug, or are the fixtures misleading us here?

project/tagging/tests.py
```diff
@@ -1,3 +1,44 @@
...
+
+            {"name": "❤", "expected_slug": ""},
+            {"name": "🐸", "expected_slug": "_1"},
```
chris48s: In these two cases, the tests pass based on the values from the fixtures but it seems like these aren't generating particularly useful slugs (esp when you consider the objective of slugging is to make the text unique and URL-safe). What should we be doing here?
If a single emoji isn't a valid tag name, shoudl we just be validating here so that trying to save a tag called "🐸" throws `ValidationError`?

project/tagging/tests.py
BethanyG: So.  Two different bugs here, in different places.  Both are concerning/irritating.  

1.  The code is indeed returning **"-जनकर"** as the slug of **"हिंदी में जानकारी"** which is entirely incorrect.  In fact, **"-जनकर"** isn't a word in Hindi.  Grr.  There appears to be some hiccough in **`slugify`** that deals with spaces in Hindi words very poorly.  Additionally, the upper-case to lower-case character mapping appears to be way off.  If I attempt to slugify **"हिंदी-में-जानकारी"** ("information in Hindi"), I get back **"हद-म-जनकर"** ("half-dead"). A similar thing seems to be happening with Telugu - **"స్వయంచాలక"** ("automated") becomes **"సవయచలక"** -- which isn't a word.  However, slugifying **"базы данных"** ("database") results in the expected **"базы-данных"**.  Ditto for **"قاعدة البيانات"** ("Database"), which slugifies to **"قاعدة-البيانات"** ("Database")  _sigh_.

2.  But irrespective of slugify and it's mangling of Indic languages, we seem to have a problem with Pytest/unittest.  Because those are accurate strings for what the code is returning, so there shouldn't be a failure in matching .. unless pytest is not reading the UTF-8 correctly.

Will have to dig more..but wanted to comment on what I've found so far.

project/tagging/tests.py
BethanyG: I think that if we can't get emojis to work correctly through slugifying, then yes - we should throw a validation error.  I can log an issue about that...along with our issues around Hindi and Telugu.  I really don't want to have to transliterate, but we may need to go that route if we can't find an alternative to what slugify is doing to those languages.

project/tagging/tests.py
BethanyG: Ugh.  CORRECTION:  The fixture **is wrong**, and loading the fixture then screws up things down the road when you try and add variants...because _uniqueness_. 🙄 

Still doesn't change the fact that the slugify code is mangling Brahmic scripts .. but I've logged that already.  For the 4 tests that aren't passing (for now -- until we fix the mangling issue), we just need to change the expected result to the one returned directly by `slugify()` in the Django shell. Will push that shortly.

Will also update the fixture file with these examples.

BethanyG: Ok.  Did a commit directly here for the test fixes....wasn't exactly sure how to do it to your branch.  Oof.  Apologies.  Will dump the fixture fix into ... some subsequent PR.

project/tagging/tests.py
BethanyG: Logged issue #123 for the script fail and #124 for the emoji.

BethanyG:
Going to link issue #72 here as well, since this PR does address some of those scenarios.

project/tagging/tests.py
lpatmo: (fwiw from a product perspective, I think it's totally fine for us not to support emojis in tags. Validating against it sounds reasonable!)

project/tagging/tests.py
chris48s: Just to zoom out slightly here.. can you explain a bit more about how you're planning to use the slugs in the application?
Is it used by the front-end? If so, how?
Are you planning to query the API by slug or GUID?
I think understanding your objective would be helpful at this point.

BethanyG:
@chris48s @lpatmo - I think we can probably go ahead and merge this.  We'll need to do another pass, once we get solutions/strategies for #123, #124 with related tests .. but I think this is decent start, at least.

BethanyG:
@lpatmo  - since I contributed to this as well, I'd feel better if you gave it a quick review.  😄 Thanks!

lpatmo:
Just took a look -- thank you for writing this and it looks good, @chris48s! Feel free to merge in!

chris48s:
Its been a while since I looked at this, but I don't think this is finished.

I'd like to at least add a validator to throw `ValidationError` on tags with emoji and test that, rather than formalising that a single emoji tag generates a blank slug if that's the desired behaviour.

That said, there are still some open questions here: https://github.com/codebuddies/backend/pull/121#discussion_r400403287

I'd like to understand more about the use-case for slugs in this application first, but if you take https://github.com/codebuddies/backend/issues/123 at face value, c320bc7 essentially formalises a bug, doesn't it? i.e: the tests pass because the code _doesn't_ do what you want it to. That's the opposite of what you want from a test.

Either way, lets understand what you're doing with the slugs first..

project/tagging/tests.py
lpatmo: Taking a stab at these questions...

> Is it used by the front-end
Yep, used by the front-end as part of URL slugs. So when a user clicks on a tag, they should be able to go to something like codebuddies.org/tags/javascript and see the resources/hangouts/learnings tagged "javascript."

> Are you planning to query the API by slug or GUID?
Hmm... I think by slug instead of by the GUID. This will happen when a user is type-searching for a slug to add to a particular resource. 

fwiw I'm not anticipating that emoji or non-English slugs at the moment, so while those are good things to file, I'm not too worried about them right now.

lpatmo:
I refreshed the page and just saw your new comment! I took a stab at answering the questions in #121 (comment) just now. :)

lpatmo:
> I'd like to at least add a validator to throw ValidationError on tags with emoji and test that

++

project/tagging/tests.py
chris48s: > Yep, used by the front-end as part of URL

Cool, so the key thing you don't want to break is URL-safety :+1:

> This will happen when a user is type-searching for a slug to add to a particular resource

Surely the human-facing representation of the tag (and the thing you search on) is the name, not the slug?

lpatmo: > Surely the human-facing representation of the tag (and the thing you search on) is the name, not the slug?

Ahh you're right, it _is_ the name! I can't think of any cases where we'd query the API by slug or GUID. 🤔

project/tagging/tests.py
BethanyG: So - I think adding in validators  (and tossing errors) for emoji and picto-ascii  would be a good thing.  But we have some behaviors to think through for the api/app.   I am seeing multiple scenarios - but also have questions:

1.  Single or repeated unicode or ascii emoji/symbols as the sole content of a tag :  reject out of hand.
     *  validator for DRF should toss an error here.  HOWEVER - does that mean the entire creation of the `resource` fails?  I don't think so, since tags are not _technically_ part  of a `resource`....but that then creates some issues.
    *  if a validation error **is** thrown, do we create the `resource`, but hand back a message to the effect that one or more of the `tags` was invalid?  HTML [**207**](https://httpstatuses.com/207) with an embedded 400 and a message of _"Your tag text contains one or more unsupported characters, and was not saved."_?  This needs some detailed scoping.
    *  if a validation error is thrown, do we create the `resource` but omit all the tags and hand back a message?
    *  if only **one** of multiple tags has a problem, do we drop the "bad" tag, or do we fail the whole set of tags?

2.  Single or repeated unicode or ascii emoji/symbols as _**part**_ of a tag :  do we drop these silently - creating the tag/slug without them, or do we throw a validation error for the entire tag?  And if we do throw a validation error for the entire tag, which scenario from above do we apply when creating the `resource` ?

3.  When we validate, will we be validating for both the name and the slug?

project/tagging/tests.py
BethanyG: The slug here is primarily for uniquification.  If we were searching,  we'd probably be doing some sort of similarity search across all the tags we could net for the term.  For example:  javascript would include JS, Java Script, java script, js, and quite possibly ES6 or some other terms, depending on how we implemented it.  So "clicking on the tag" would most likely not have a 1-to-1 relationship with either the name or the slug...but we **do** want a methodology of making our massive pile of tags a bit less massive/repetitive.

As for the GUID ....  I don't see a user-facing need for it.  Doesn't mean I am going to delete it from the model, but I am not seeing it as part of a URL right now.  Although if we follow our `resource` pattern, going to the GUID of a tag would let you edit that specific tag...which doesn't feel like something we want.

BethanyG:
> I'd like to understand more about the use-case for slugs in this application first, but if you take #123 at face value, c320bc7 essentially formalises a bug, doesn't it? i.e: the tests pass because the code doesn't do what you want it to. That's the opposite of what you want from a test.

Yup - technically speaking, it does.  So - we can re-write those tests to fail .. but then we need to decide what the expected behavior is...and write some slugification code (or [**iri_to_uri code**](https://docs.djangoproject.com/en/3.0/ref/unicode/#uri-and-iri-handling)) accordingly...or live with 4 continuously failing tests (which seems silly).  We could also omit them altogether until we decide.

chris48s:
OK then. Based on the further discussion, it seems like one of the things you want from slugging is to enforce uniqueness i.e: you shouldn't be able to create `'javascript'`, `'Javascript'` as 2 seperate tags. With the code as it stands on `master`, I can happily create these as different tags. For example:

```py
>>> tag = CustomTag(name='javascript')
>>> tag.save()
>>> tag.slug
'javascript'

>>> tag = CustomTag(name='Javascript')
>>> tag.save()
>>> tag.slug
'javascript_1'

>>> tag = CustomTag(name='JavaScript')
>>> tag.save()
>>> tag.slug
'javascript_2'
```

That seems like it is not what you want, so in cdfe87a I have changed this so that:

- we can't have 2 tags which generate the same slug
- we can't have any tag which generates an empty slug (e.g: no single emjoi tag names, or tags called "%" or whatever)

and I've added tests to formalise that behaviour.


On balance, that seems more like what you want. This does, of course, create at least one annoying edge case to work around:

```py
>>> from django.utils.text import slugify
>>> slugify('C')
'c'

>>> slugify('C++')
'c'

>>> slugify('C#')
'c'
```

That's what I've don so far, but it looks like there is also some further validation needed beyond this. Lets carry on that discussion in https://github.com/codebuddies/backend/pull/121#discussion_r403644974

---

> So - we can re-write those tests to fail .. but then we need to decide what the expected behaviour is...and write some slugification code (or iri_to_uri code) accordingly...or live with 4 continuously failing tests (which seems silly). We could also omit them altogether until we decide.

My proposed solution is 49eca65 - we move those examples to a skipped test. This means:
- There is a record in the test suite of the problematic inputs when someone wants to look at it.
- The build passes.
- We don't have to deal with fixing it _right now_.
- We haven't written a test that passes because of behaviour you don't want.

btw, I'm slightly confused on the `stòran-dàta` example. Was that one just a typo in the fixture, or is it a "special" `s` that should actually slug to `ss` (like a german `ß`)? Should just one just be moved back to `test_valid_slugs()` as `{"name": "stòran-dàta", "expected_slug": "stòran-dàta"}`?

project/tagging/tests.py
chris48s: > Single or repeated unicode or ascii emoji/symbols as the sole content of a tag : reject out of hand.

Done.

> Single or repeated unicode or ascii emoji/symbols as part of a tag : do we drop these silently - creating the tag/slug without them, or do we throw a validation error for the entire tag?
>  When we validate, will we be validating for both the name and the slug?

So far, the only validation I've done applies to the _slug_, so basically whichever variation claims the slug first gets it i.e: "Javascript 🙂" is a legal tag name, and if you create "Javascript 🙂" first, that tag now owns the slug `javascript`.

If you want to apply the same rules to the name, I wonder if it is actually useful for the tag name and slug to be different?

---

The other questions on API behaviour are probably best moved to another issue to avoid trying to boil the ocean in one PR as this is already getting long. Lets just focus on the model behaviour here.

My gut instinct though is that working out these behaviours would probably be much easier if the operations of:

- creating tags
- creating resources
- applying tags to resources

were 3 separate endpoints/operations. Then you can know if your tag is valid before you apply it to a resource. You can still tie those 3 things into what feels like a single 'page' on the frontend for a nice UX.

lpatmo: Ohh, I see -- so after a user finishes typing a tag, we can fire off a POST request to a /tags endpoint to check whether that tag is valid. _Then_ they submit the form.

> applying tags to resources

Can you say more about how this endpoint would work?

---
> HOWEVER - does that mean the entire creation of the resource fails?

From a naive user perspective, if I was submitting to create a resource and had a tag that was invalid in some way (e.g. had an emoji character), then having the entire form submission fail on submit (e.g. no tags or resources is created) is reasonable to me. I'd expect to see an error about some invalid character in one of the tags, change it appropriately, then re-submit the form.

BethanyG: Humm.. as @chris48s suggested -- we probably should take this to a separate issue.  So I will make that  shortly.  Especially since this convo will be hidden once the PR is merged in.  

I think the web UX could be fairly straightforward, and could include both a UI message (_**the following characters are not allowed for tags**_) &  blocking the user from typing said emoji/ascii/character sets (a background form/field validation that dropped or highlighted any unicode or ascii characters in that range, similar to what I have seen with some set password fields).  I don't think there is even a need there to fire off a POST - we just need to agree which characters/character sets are going to be off limits, and then **fail** them before they can get to the backend (theoretically).

But the API itself is a bit different (since we're building for more than the website, and can't give feedback to users in the same way) -- and in that case, for clarity, we probably do want  to have a whole creation failure of an object with a message about why.  (_**"your tags contain invalid/out of bounds characters.  Characters in the following ranges are not allowed. ..."**_).  

OR we have a a separate tagging endpoint & action (_pass in the GUID of the object to be tagged, along with a list of tags and get back a verification that tagging for that object has succeeded or failed w/ a message about why_).  

Now - **what** characters would cause that is still a bit open....

I would disallow emoji and picto ascii altogether in either tag **names** or **slugs**.  Why go there?  Doesn't seem to add significant value to our users at the moment. 

On the other hand, having the ability to tag or note something in your native language **does** have value - so I still want to noodle on that problem.

BethanyG: I am going to copy this thread into a new issue for discussion/add to an existing issue....

BethanyG:
> btw, I'm slightly confused on the stòran-dàta example. Was that one just a typo in the fixture, or is it a "special" s that should actually slug to ss (like a german ß)? Should just one just be moved back to test_valid_slugs() as {"name": "stòran-dàta", "expected_slug": "stòran-dàta"}?

Apologies, @chris48s - missed this in the whole comment shuffle.  **YES** - AFAK it was a typo, (I don't think  Scots Gaelic has an Eszett-type character..and Google appears to agree with me.) so let's just move `{"name": "stòran-dàta", "expected_slug": "stòran-dàta"}` back into test data.

project/tagging/tests.py
lpatmo: Sounds good! (Btw, to clarify my earlier statement, I was talking about the error message from an API call as well -- that is, even though we'd have front-end validation, I'd like to see an error message returned if I say make a POST request in Postman to create a resource with problematic tags included. Then I'd expect the entire request to fail, if that makes sense. :P) 

image
--> error

BethanyG: Move this over to issue #127.

chris48s:
I've marked this PR as ready for review now. There are still quite a few unanswered questions about what should/shouldn't be allowed in a tag name/slug but I don't think we've reached a conclusion that results in clearly defined changes to this PR yet.

My suggestion would be we review/merge this as it stands now and revisit following further discussion in issues. It may be that loads of this stuff ends up needing to change soon, but if nothing else this PR forces us to nail down some decisions about stuff that was previously not very well defined :)

PR #126, opened 2020-04-05
Re-wrote the existing Resources tests to use resources.factories.

BethanyG:
Ok.  Re-wrote the existing `resources` tests (and found a new bug!).  I think this is ready to review and merge.  Will open an issue for additional tests that you or Mar (or I) can work on....but let's close this one (assuming all is well with my tests....) out.


BethanyG:
arg.  On second thought?  Will push to the original PR.  Nevermind.

PR #129, opened 2020-04-08
Generate seed data for dev/staging using factories

### LINKED ISSUE #115: [Testing] [WIP] Use Factory-Boy to Create  Seed Data for Dev and Staging DBs

chris48s:
Closes #115

This PR adds a django management command `init_data` which can be used to populate the DB with some test data and removes the existing JSON fixtures.
The test data generated is random and meaningless, so you'll end up with a resource called "Road stage could least" tagged with "blue" and "head", but as the models and tests evolve the test data we can generate should evolve with it. It also gives us flexibility about how much data we generate, for example.
It might be that there are particular conditions we want to ensure are always created in the test data. For example, one I've already ensured happens is that at least one resource has zero tags attached to it, but if we think of more we can always extend this, and it should be much easier than manipulating ~1000 line JSON files.

```
$ docker-compose run --rm app ./manage.py init_data --help

usage: manage.py init_data [-h] [--clear-db] [--num-users NUM-USERS]
                           [--num-tags NUM-TAGS]
                           [--num-resources NUM-RESOURCES] [--version]
                           [-v {0,1,2,3}] [--settings SETTINGS]
                           [--pythonpath PYTHONPATH] [--traceback]
                           [--no-color] [--force-color]

Initialize the DB with some random fake data for testing and development

optional arguments:
  --clear-db            Clear existing data from the DB before creating test
                        data
  --num-users NUM-USERS
                        Number of `User` objects to create (default 10)
  --num-tags NUM-TAGS   Number of `Tag` objects to create (default 10)
  --num-resources NUM-RESOURCES
                        Number of `Resource` objects to create (default 10)
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Verbosity level; 0=minimal output, 1=normal output,
                        2=verbose output, 3=very verbose output
```


project/core/management/commands/init_data.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def teardown(self):
+        TaggedItems.objects.all().delete()
+        CustomTag.objects.all().delete()
+        Resource.objects.all().delete()
+        User.objects.all().delete()
+
+    def create_users(self, num):
+        # Create random users.
+        # All users have the password 'codebuddies'.
```
lpatmo, line 68: This is fancy :D And all those possible arguments! 😍 

One thing I noticed: I logged into adminer to check the new users that were created, and saw:

image

So I tried authenticating via one of those usernames and the password `codebuddies`, but was not able to do so. 

image

chris48s: Thanks - well spotted. I've fixed this in dcca264 I'm not 100% sure why this fixes it. Maybe `create_batch()` doesn't fire the `post_generation` hook or something? Either way, its fixed..
I didn't notice this because I was originally creating users with a loop and then switched them to a `create_batch()` later :roll_eyes:

lpatmo: (Sorry for slow reply -- thanks for fixing! :))

Issue #125, opened 2020-04-05
[Resources] Media Type Choices Field Value is Dropped when POSTing in JSON


BethanyG:
_This is why we test._  🙄 

If a `resources` record is created via POST (using cURL, Postman, charles, or other client), the value in the `media_type` field (defined as a CHOICES field) is dropped.  It is handled OK if the POST is done via the browsable api (which uses a Django form view), but needs to have an additional methods for the DRF serializer to render & save it correctly.  

See [**this Stack Overflow Post**](https://stackoverflow.com/questions/28945327/django-rest-framework-with-choicefield)  & [**this additional Stack Overflow Post**](https://stackoverflow.com/questions/48087416/django-rest-framework-choices-field-serializer) for details on how to fix this.

Essentially, we need to add some variant of:

```python

 media_type = serializers.SerializerMethodField('get_media_type')

# ......

    def get_media_type(self, obj):
        return obj.get_media_type_display()

    def to_internal_value(self, data):
        return data
```

To `resources/serializers.py`, so that the internal value is saved, and the "display name" is properly fetched for the return serialization/representation.

Until this is done and checked in, one of the `resources` tests (`test_create_one_resource`) will fail.

PR #134, opened 2020-04-16
Make the GET /resources endpoint public, and update tests

lpatmo:
## Context
The GET /resources and a GET /resources/{id} required user authorization. But ideally, we want non-logged-in-users to be able to search and look at resources. They should only have to be logged in if they want to delete, create, or edit a resource.

@chris48s and @BethanyG did most of the heavy lifting with tests already; I only followed the same pattern.

Feedback welcome!

Note: I left POSTing a resource as a later todo, since it is blocked by https://github.com/codebuddies/backend/issues/125. 

## Before & After

Before:
GET /resources and GET /resources/{id} hidden under a token

After:
GET /resources and GET /resources/{id} available to users without a token

## Implementation
[x] Make GET /resources not protected by authorization
[x] Change setup for tests to make sure we're not authed for the GET requests, but are authed for PATCH, and search.
[x] Add test to make sure GET /api/v1/resources/{{guid}}/ and GET /api/v1/resources/ requests and search work _without_ a token
[x] Add/alter tests for GET GET /api/v1/resources/{{guid}}/ and GET /api/v1/resources/ requests and search to ensure that they also work _with_ a token
[x] Add tests to make sure PATCH and POST fail without a token
[x] Add test to make sure DEL fails without a token


lpatmo:
> Personally I find reducing each test case to the bare minimum makes it easier to read/understand

That is excellent feedback, thank you! Will take a look/improve this tonight...

lpatmo:
Pushed up some simplifying changes in 2caac24!

lpatmo:
@chris48s Thanks!!

PR #140, opened 2020-05-02
[Issue 125 Fix] :  Added Serializer Field to Properly Resolve Choices in Resource.media_type Field

### LINKED ISSUE #125: [Resources] Media Type Choices Field Value is Dropped when POSTing in JSON

BethanyG:
Should fix #125.

*  Added `class MediaTypeSerializerField(serializers.ChoiceField)` with a `to_representation` and a `to_internal_value` to properly resolve the `choices` on the `Resource.media_type` field for Resources. in **`resources/serializers.py`**

*  Corrected and un-commented the `create_one_resource_with_media_type` test case and added a new `create_one_resource_without_media_type` test case for coverage. in **`resources/tests.py`**

### DIFF
--- a/project/resources/serializers.py
+++ b/project/resources/serializers.py
@@ -7,7 +7,20 @@ from tagging.serializers import TagSerializer, TagsSerializerField
 class MediaTypeSerializerField(serializers.ChoiceField):
 
     def to_representation(self, value):
-            return "" if not value else self.choices[value]
+
+        valid_media_types = ', '.join(item for item in self.choices)
+
+        if not value:
+            return ''
+
+        else:
+            try:
+                media_type = self.choices[value]
+
+            except KeyError as err:
+                raise KeyError(f'Invalid media type.  The media type should be one of the following: {valid_media_types}') from err
+
+            return media_type
 
     def to_internal_value(self,  value):
         return value
--- a/project/resources/serializers.py
+++ b/project/resources/serializers.py
@@ -4,11 +4,19 @@ from userauth.serializers import UserSerializer
 from tagging.serializers import TagSerializer, TagsSerializerField
 
 
+class MediaTypeSerializerField(serializers.ChoiceField):
+
+    def to_representation(self, value):
+            return "" if not value else self.choices[value]
+
+    def to_internal_value(self,  value):
+        return value
+
 
 class ResourceSerializer(TagSerializer, serializers.ModelSerializer):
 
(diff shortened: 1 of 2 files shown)


project/resources/serializers.py
```diff
@@ -4,11 +4,32 @@
...
+        valid_media_types = ', '.join(item for item in self.choices)
+
+        if not value:
+            return ''
+
+        else:
+            try:
+                media_type = self.choices[value]
+
+            except KeyError as err:
```
lpatmo, line 20: One question here: when I passed in an invalid `media_type` field, I was hoping that the key error would say "Invalid media type. The media type should be one of the following: [list]`, but instead the error from Postman is more vague: 
image

Do you know why?

project/resources/tests.py
```diff
@@ -1,4 +1,5 @@
 from unittest import skip
+from pytest import raises
```
lpatmo, line 2: Nice, didn't know about `pytest.raises` 👍

project/resources/serializers.py
BethanyG: _interesting_.  It did so for me, but I also POSTED a full record:

image

lpatmo: I see that now, after pulling in this merged branch in `master`! I wonder if maybe I needed to do `docker-compose down` and re-run it 🤷Anyway, solved now :)

PR #141, opened 2020-05-18
Issue 124 (as well as 123 & 127) - Tagging Tests, Fixes & Touchup

### LINKED ISSUE #123: [Tagging] Slugify Mangling Tags Written in Brahmic scripts / abugida writing systems (Vowels are being stripped)
The expectation of the `slugify(allow_unicode=True)` call in the `tagging/models.py` **`CustomTag()`** class is to pass any tag written in unicode characters through mostly untouched except for 
1.  removal of leading and trailing spaces
2.  replacing spaces between words with the **-** character
3.  lower-casing the words
4. ~making sure the resulting tag is "unique"~ 
   (_this is actually the job of the taggit manager code_)

_________________________

However, when tags are written in a Brahmic / abugida writing system (examples include Hindi,  Telugu, Thai, Malayalam, Tamil, Kannada, and more) this code is mangling the result by removing the diacritical marks and vowels.  Going off the Google translations here, as I am not a speaker of any of the example languages.

The slug of **"हिंदी में जानकारी"** ("Information in Hindi") is being returned as  **"-जनकर"** which isn't a word.  Attempting to then slugify **"हिंदी-में-जानकारी"** ("information-in-Hindi"), I get back **"हद-म-जनकर"** ("half-dead").   

A similar thing seems to be happening with the Telugu language - **"స్వయంచాలక"** ("automated") becomes **"సవయచలక"** -- which isn't a word. 

Additional examples:

Kannada:    **"ಡೇಟಾಬೇಸ್ ನಿರ್ವಹಣಾ ವ್ಯವಸ್ಥೆ"** becomes  **"ಡಟಬಸ-ನರವಹಣ-ವಯವಸಥ"**
Malayalam: **"ഡാറ്റാബേസ് മാനേജുമെന്റ് സിസ്റ്റം"** becomes  **"ഡററബസ-മനജമനറ-സസററ"**
Thai:            **"ฐานข้อมูล"** becomes **"ฐานขอมล"**
Burmese:    **"ဒေတာဘေ့စစီမံခန့်ခွဲမှုစနစ်"** becomes **"ဒတဘစစမခနခမစနစ"**

The real kicker here is that none of these languages really have a "lower case" vs "upper case" distinction, really.

However, slugifying  the Hebrew "מערכת ניהול מסדי נתונים" ("database management system") results in the expected "מערכת-ניהול-מסדי-נתונים". 

And slugifying the Arabic "قاعدة البيانات" ("Database"), results in  "قاعدة-البيانات" ("Database").

Tests with traditional & simplified Chinese characters, Korean, and multiple Japanese variants are also fine, as is Persian.

We may need to move to either a combination of slugify with unicode and transliteration, do a run-around of slugify for certain languages - or scrap slugification altogether.  Very open to suggestions or discussions on this.

### LINKED ISSUE #124: [Tagging] Slug-ification of  Emoji & ASCII Images Results in Useless/Unidentifiable String
While emoji can be successfully saved to postgreSQL, they cannot be successfully or safely passed through Djangos `slugify()` code - while emoji **can** be used as part of a URL  see [this](http://factmyth.com/factoids/you-%F0%9F%91%8F-can-%F0%9F%91%8F-put-%F0%9F%91%8F-emojis-%F0%9F%91%8F-in-%F0%9F%91%8F-urls-%F0%9F%91%8F/) obnoxious example...it is still frowned upon, and Djangos code doesn't tolerate it.

If a tag is entered that contains only emoji, Django assigns either blank space, or an underscore plus a number to represent the "slug" of the string of characters:

image

Emoji inside a string is simply dropped:

image

Similar problems arise with ASCII image characters:

image

Because of this, we need to put validation code around tags, and fail tag strings that include emoji or other ASCII / Unicode "image" characters.

@lpatmo  - we definitely need to do this for the backend, and probably should also do this around form/entry  validation on the front end.

### LINKED ISSUE #127: [Tagging] Implementation of Tag Validation
##  Background

While implementing tagging tests, a number of bugs with the `slugify()` code and tagging were revealed.  In particular, issue #123 and #124 have started a discussion around what we want to allow or disallow in tags, how we want validation and error handling to work with tagging, and weather or not we want to include related tags from a particular endpoint (e.g. resources, hangouts, learning paths, discussions, projects, etc.).   Since much of that discussion happened inside the PR for tagging tests, we decided we needed to port it to it's own issue, as we work out the intended behavior and archetecture.

Relevant section of  PR #121  around the implementation of tag validation is below:

______

#### [chris48s](https://github.com/chris48s)** [7 days ago](https://github.com/codebuddies/backend/pull/121#discussion_r399795637) 

Author Member

In these two cases, the tests pass based on the values from the fixtures but it seems like these aren't generating particularly useful slugs (esp when you consider the objective of slugging is to make the text unique and URL-safe). What should we be doing here?
If a single emoji isn't a valid tag name, shoudl we just be validating here so that trying to save a tag called "🐸" throws `ValidationError`?

#### **[BethanyG](https://github.com/BethanyG)** [7 days ago](https://github.com/codebuddies/backend/pull/121#discussion_r399869794) 

Member

I think that if we can't get emojis to work correctly through slugifying, then yes - we should throw a validation error. I can log an issue about that...along with our issues around Hindi and Telugu. I really don't want to have to transliterate, but we may need to go that route if we can't find an alternative to what slugify is doing to those languages.

#### **[BethanyG](https://github.com/BethanyG)** [7 days ago](https://github.com/codebuddies/backend/pull/121#discussion_r399926078) 

Member

Logged issue [#123](https://github.com/codebuddies/backend/issues/123) for the script fail and [#124](https://github.com/codebuddies/backend/issues/124) for the emoji.

#### **[lpatmo](https://github.com/lpatmo)** [7 days ago](https://github.com/codebuddies/backend/pull/121#discussion_r399974490) 

Member

(fwiw from a product perspective, I think it's totally fine for us not to support emojis in tags. Validating against it sounds reasonable!)

👍 1

#### **[BethanyG](https://github.com/BethanyG)** [20 hours ago](https://github.com/codebuddies/backend/pull/121#discussion_r403644974) 

Member

So - I think adding in validators (and tossing errors) for emoji and picto-ascii would be a good thing. But we have some behaviors to think through for the api/app. I am seeing multiple scenarios - but also have questions:

1. Single or repeated unicode or ascii emoji/symbols as the sole content of a tag : reject out of hand.
   - validator for DRF should toss an error here. HOWEVER - does that mean the entire creation of the `resource` fails? I don't think so, since tags are not *technically* part of a `resource`....but that then creates some issues.
   - if a validation error **is** thrown, do we create the `resource`, but hand back a message to the effect that one or more of the `tags` was invalid? HTML [**207**](https://httpstatuses.com/207) with an embedded 400 and a message of *"Your tag text contains one or more unsupported characters, and was not saved."*? This needs some detailed scoping.
   - if a validation error is thrown, do we create the `resource` but omit all the tags and hand back a message?
   - if only **one** of multiple tags has a problem, do we drop the "bad" tag, or do we fail the whole set of tags?
2. Single or repeated unicode or ascii emoji/symbols as ***part*** of a tag : do we drop these silently - creating the tag/slug without them, or do we throw a validation error for the entire tag? And if we do throw a validation error for the entire tag, which scenario from above do we apply when creating the `resource` ?
3. When we validate, will we be validating for both the name and the slug?

#### **[chris48s](https://github.com/chris48s)** [12 hours ago](https://github.com/codebuddies/backend/pull/121#discussion_r403695805) 

Author Member

> Single or repeated unicode or ascii emoji/symbols as the sole content of a tag : reject out of hand.

Done.

> Single or repeated unicode or ascii emoji/symbols as part of a tag : do we drop these silently - creating the tag/slug without them, or do we throw a validation error for the entire tag?
> When we validate, will we be validating for both the name and the slug?

So far, the only validation I've done applies to the *slug*, so basically whichever variation claims the slug first gets it i.e: "Javascript 🙂" is a legal tag name, and if you create "Javascript 🙂" first, that tag now owns the slug `javascript`.

If you want to apply the same rules to the name, I wonder if it is actually useful for the tag name and slug to be different?

------

The other questions on API behaviour are probably best moved to another issue to avoid trying to boil the ocean in one PR as this is already getting long. Lets just focus on the model behaviour here.

My gut instinct though is that working out these behaviours would probably be much easier if the operations of:

- creating tags
- creating resources
- applying tags to resources

were 3 separate endpoints/operations. Then you can know if your tag is valid before you apply it to a resource. You can still tie those 3 things into what feels like a single 'page' on the frontend for a nice UX.

#### **[lpatmo](https://github.com/lpatmo)** [3 hours ago](https://github.com/codebuddies/backend/pull/121#discussion_r403755601) 

Member

Ohh, I see -- so after a user finishes typing a tag, we can fire off a POST request to a /tags endpoint to check whether that tag is valid. *Then* they submit the form.

> applying tags to resources

Can you say more about how this endpoint would work?

------

> HOWEVER - does that mean the entire creation of the resource fails?

From a naive user perspective, if I was submitting to create a resource and had a tag that was invalid in some way (e.g. had an emoji character), then having the entire form submission fail on submit (e.g. no tags or resources is created) is reasonable to me. I'd expect to see an error about some invalid character in one of the tags, change it appropriately, then re-submit the form.

#### **[BethanyG](https://github.com/BethanyG)** (https://github.com/codebuddies/backend/pull/121#discussion_r403776601) 

Member

Humm.. as [@chris48s](https://github.com/chris48s) suggested -- we probably should take this to a separate issue. So I will make that shortly. Especially since this convo will be hidden once the PR is merged in.

I think the web UX could be fairly straightforward, and could include both a UI message (***the following characters are not allowed for tags***) & blocking the user from typing said emoji/ascii/character sets (a background form/field validation that dropped or highlighted any unicode or ascii characters in that range, similar to what I have seen with some set password fields). I don't think there is even a need there to fire off a POST - we just need to agree which characters/character sets are going to be off limits, and then **fail** them before they can get to the backend (theoretically).

But the API itself is a bit different (since we're building for more than the website, and can't give feedback to users in the same way) -- and in that case, for clarity, we probably do want to have a whole creation failure of an object with a message about why. (***"your tags contain invalid/out of bounds characters. Characters in the following ranges are not allowed. ..."***).

OR we have a a separate tagging endpoint & action (*pass in the GUID of the object to be tagged, along with a list of tags and get back a verification that tagging for that object has succeeded or failed w/ a message about why*).

Now - **what** characters would cause that is still a bit open....

I would disallow emoji and picto ascii altogether in either tag **names** or **slugs**. Why go there? Doesn't seem to add significant value to our users at the moment.

On the other hand, having the ability to tag or note something in your native language **does** have value - so I still want to noodle on that problem.

#### lpatmo](https://github.com/lpatmo)** [22 minutes ago](https://github.com/codebuddies/backend/pull/121#discussion_r403802305) 

Member

Sounds good! (Btw, to clarify my earlier statement, I was talking about the error message from an API call as well -- that is, even though we'd have front-end validation, I'd like to see an error message returned if I say make a POST request in Postman to create a resource with problematic tags included. Then I'd expect the entire request to fail, if that makes sense. :P)

[image](https://user-images.githubusercontent.com/4512699/78518311-0101de00-7775-11ea-9a05-d918f3a5b138.png)
--> error

_____

##  Decisions We've Made So Far:

1.  The backend/api will throw a `ValdiationError` for any `tag` that contains what we consider "invalid" characters.
2.  A validation error will cause the POST request to create a new `resource` to fail with a return message along the lines of "One or more of your tags contain invalid/unsupported characters."

##  What We Still Need to Decide/Discuss

1.  What sets of characters will cause a `ValidationError`?
2.  Will both the **name** of the tag and the **slug** of a tag be validated?
3.  Will we continue having tags returned as part of a `resources` endpoint, or will we move toward having `tagging` as its own endpoint?
4.  If `tagging ` becomes its own endpoint, what does that look like, and what is the expected flow/interaction?

BethanyG:
### Issues 
Closes #123 - Mangled Abugida slugs
Closes #124-  Emoji and Pitctograms slugging to blank spaces. 
Closes #127-  Tag Validation 
____

**Please see the following two github threads for some details on how `taggit` handles situations where a `tag.name` and a `tag.slug` match, and the app is set to `TAGGIT_CASE_INSENSITIVE=True`**:

1.  https://github.com/jazzband/django-taggit/issues/448#issuecomment-414474054
2. https://github.com/wagtail/wagtail/issues/4786#issuecomment-426436030

_____

### Changes

Decided to remove the unique constraint from the slug field of `CustomTag`.  Now, if two tags slug to the same value but their `name/slug` combo is _unique_, the tag name is saved to the DB, and the slug value is repeated (_as in the slug is no longer unique in its column_).  What this means is that queries for tag slugs will return more than one name value...so we will have to account for that in search.

Also decided to bypass Djangos built-in `slugify` for Abugida scripts.  Now, if an Abugida is detected, it will go through a different slugification process that does not remove the combination and diacritical marks.

Finally, wrote a regex that rejects any emoji, pictograms, letter-like symbols, etc.  Any of these will now throw a `serializer.ValidationError` - as will items that are not of type `str()` and items that are not in a `list`.

Tests are ongoing, and I probably missed some scenarios (_feedback and help always welcome!_).  Some of them were more involved than others, since validation code is only triggered through the serializers, and does not get called by invoking `tag.save()`.  More duplicate and unique_together tag tests are probably needed as well.


project/tagging/tests.py
```diff
@@ -1,77 +1,208 @@
...
             {"name": "אַלגערידאַם", "expected_slug": "אלגערידאם"},
             {"name": "自動化する", "expected_slug": "自動化する"},
+            {"name":"מאגר מידע ניהול מערכת", "expected_slug":"מאגר-מידע-ניהול-מערכת"},
             {"name": "sjálfvirkan", "expected_slug": "sjálfvirkan"},
             {"name": "پژوهش ", "expected_slug": "پژوهش"},
-            {"name": " గ్రాఫ్", "expected_slug": "గరఫ"},
+            {"name": " గ్రాఫ్", "expected_slug": "గ్రాఫ్"},
             {"name": "데이터 베이스", "expected_slug": "데이터-베이스"},
             {"name": "stòran-dàta", "expected_slug": "stòran-dàta"},
+            {"name": "የመረጃ ቋት አስተዳደር ስርዓት", "expected_slug": "የመረጃ-ቋት-አስተዳደር-ስርዓት"}
```
lpatmo, line 50: Impressed with all of these language examples.

lpatmo:
This looks great! I was having a bit of trouble getting the branch to run without displaying an nginx error, though -- it looked like it was caused by this: 

```
ModuleNotFoundError: No module named 'regex'
```

So I ran `docker-compose run --rm app pip install -r requirements/base.txt` and I thought I saw regex was installed, but am still seeing this:

image

Wonder if there's another pip install command I'm missing 🤔

lpatmo:
Thanks so much for knocking out all those issues, btw :) Once I get my branch to run I'll do a bit more Postman testing!

BethanyG:
So...I am still not totally clear on all the intermediary build steps, but here are three strategies that worked for me when I needed to get the `regex` library + associated changes to the DB installed & working on my docker/system combo:

1.  During dev, in order to try out the library before making changes to `requirements/base.txt`:
     *  **while the app container was running**, `docker exec -it app /bin/bash` (_to log into the app container with an interactive bash shell_)
     *  At the `/opt/codebuddies` directory (_you should already be there after running the above command_), run `pip3 install` **`<insert name of library here>`**

2.  During dev, after the changes to `requirements/base.txt`, to test that things were being installed and picked up:
     *  Stop all containers via `CTRL-C`.  Follow up with `docker-compose down`
     *  Re-build all containers via `docker-compose up -d --build`

3.  During dev, after changing the `tagging` model (this supposes that the migration files have been made) and during the writing of validation code and tests:
     *  Stop all containers via `CTRL-C`.  Follow up with `docker-compose down`
     *  Clean up containers via `docker-compose rm`.  Use this as an opportunity to  follow up with any docker prune commands needed to clean up hanging and orphan containers and images.
     *  Further clean up data directory by running `docker volume rm backend_db_data` (_this is so you don't have bogus or confusing data in your DB_)
     *  Re-build code from scratch using `docker-compose up -d --build`.  Then use `docker-compose up`.
     *  Make new super user by logging into the running app container via `docker exec -it app /bin/bash` and then running `python manage.py createsuperuser`.  This can also be done via `docker-compose`.  Both work.
     *  Proceed to test, knowing that the DB is entirely empty.

...You can probably get away with using strategy **2**, but you won't really be able to cleanly test the deduping and unique together code without nuking all your data, which would use most of strategy **3**.

lpatmo:
Gahhh, I can't believe I forgot about `docker-compose up -d --build` -- that did the trick. Thanks so much for all those debugging options, Bethany! I'll create a quick PR to make sure `docker-compose up -d --build` is in the contributing.md / README.md so that future me remembers I need to rebuild to install new packages 😅

PR #143, opened 2020-06-02
Feature/readme update

tgrrr:
I added some of the links to the spec docs that I had in the CB slack channel. Feel free to add any others you have during this update.

I also updated the link to PR127, it was pointing to PR129


README.md
```diff
@@ -1,90 +1,141 @@
...
+  --num-tags NUM-TAGS   Number of `Tag` objects to create (default 10)
+  --num-resources NUM-RESOURCES
+                        Number of `Resource` objects to create (default 10)
+  -v {0,1,2,3}, --verbosity {0,1,2,3}
+                        Verbosity level; 0=minimal output, 1=normal output,
+                        2=verbose output, 3=very verbose output
 ```
 
-5. Create a superuser so that you can log into http://localhost:8000/admin by running the following in your terminal: `$ docker-compose run --rm app ./manage.py createsuperuser`
+[See PR 127]
```
lpatmo, line 107: I think this is supposed to say PR 129 so it can point to https://github.com/codebuddies/backend/issues/129

bengineerdavis: How does this relate to the custom ./manage.py script and its random test data?

README.md
```diff
@@ -1,90 +1,141 @@
...
 These instructions have been used on the following operating systems.
 
-* Linux
-* Mac OS
-* Windows 10 Pro
-
-Please note that Windows 10 Home is not supported by Docker Desktop at this time.
+- Linux
+- Mac OS
+- Windows 10 Pro - Please note that Windows 10 Home is not supported by Docker Desktop at this time.
```
bengineerdavis: Can we confirm whether these commands also work for Windows 10 Pro? I know Cygwin can let us run Unix commands, or maybe now that's integrated because of the WSL?

bengineerdavis: Because I've been AFK from this project for a bit--we can't use Windows 10 Home because the Docker for Windows doesn't support it, correct?

README.md
lpatmo: `docker-compose run --rm app ./manage.py init_data` creates test data for things like users and resources, while `docker-compose run --rm app ./manage.py createsuperuser` creates a superuser that lets you log in to `/admin` and control the various data on the app. 

(Let me know if I answered your question/if that's what you were asking about!)

README.md
lpatmo: I've been AFK for a while too, but recently learned the good news that Docker for Windows 10 Home is now supported! :) https://docs.docker.com/docker-for-windows/install-windows-home/

lpatmo: We'll probably have to do one more README update PR to clean this up.

lpatmo: I don't have Windows 10 Pro, but wonder if anyone else does. (@nuageklow?)

nuageklow: i can help out testing the commands again on both Win10Pro and WSL with Ubuntu 20.04 later this week or by the end of this weekend.

lpatmo: Thank you @nuageklow!

nuageklow: just some personal experiences from work about docker using on Windows: docker commands on Windows Subsystem Linux (WSL)  and regular Windows (10 Pro) could be different as WSL is completely Linux environment. But I agree WSL is definitely a work around for Windows 10 Home user to use docker commands.

on the other hand its a bit tricky to do certain things i.e. install packages from links (at least for me - i blame ubuntu20.04 being too new) as WSL still mainly relies on Windows system at the very back. I'm still trying to figure out about it for work purpose. will test out docker commands on readme and give feedback if there's any issues.

bengineerdavis: Thanks @nuageklow! Also thank you for your comments elsewhere on this PR. Are you able to use the bash terminal for commands, or is it still relying on powershell?

nuageklow: @bengineerdavis WSL settings like setting it up relies on powershell (like an app) but it uses bash terminal

nuageklow:
quick update on the docker django management commands testing on Win10Pro and WSL - all manage.py related commands do not work on both. 

figuring out alternative commands for both OS and will update the README.md when done.

PR #146, opened 2020-07-19
Create pull_request_template.md

lpatmo:
Template for new PRs. Feedback welcome!


BethanyG:
I think a little context or instruction under each header is needed.  None of those headers are intuitive to me as to content or information.  Was this borrowed from something or from a standard? 

In particular, I think it's important to have a checklist or a prompt for users around which issues are being addressed, which discussions have been had, and which requirements or stories were taken into account.  Also any substantive changes in behavior from previous versions.

And as a side note - why was this never tagged for review?  It is very hard to know when something needs feedback if it isn't requested for review.

Issue #179, opened 2020-09-08
[API Projects] "Model" tests for /projects endpoint


BethanyG:
Issues #80 and #81  (closed PRs #170 and #169) were approved and merged without tests being written for the code.  **This was not supposed to happen.**  Tests now need to be written for those two issues/PRs before going further with this endpoint.

- [ ] Model tests need to be written & run for the model changes made in issue #80 
- [ ] Unit tests need to be written & run for the app changes made in issue #81 
- [ ] Both sets of tests need to be added to codecov report.

Adding these links, in case they're helpful:

-  [Testing Models with Django using PyTest and Factory-Boy](https://medium.com/@dwernychukjosh/testing-models-with-django-using-pytest-and-factory-boy-a2985adce7b3#:~:text=Pytest%20and%20Factory%20Boy%20make,running%20testing%20logic%20on%20them.&text=The%20django_db%20decorator%20must%20be,to%20access%20the%20testing%20models.)
-  [Real Python:  Django PyTest Fixtures](https://realpython.com/django-pytest-fixtures/)
-  [Django Stars: Testing Your App with PyTest](https://djangostars.com/blog/django-pytest-testing/) -- longer and more detailed article on PyTest in General and tips for Django testing with PyTest
-  [PyTest Docs](https://docs.pytest.org/en/stable/contents.html)
[pytest-django Docs](https://pytest-django.readthedocs.io/en/latest/)

lpatmo:
@watchtheblur @BethanyG Test makes sense for me re: API endpoints, but I don't fully understand what we would be testing for #169 and #170, which is why I didn't ask @watchtheblur to write any (I thought they weren't necessary).

#169 involves a single `startapp` command that creates a bunch of files that are standard to Django, so am not sure what we would test. 

#170 creates a model, and from my reading of [this Obey the Testing Goat chapter on database layer validation](https://www.obeythetestinggoat.com/book/chapter_database_layer_validation.html) I understood that it makes sense to test that validation works as intended, but the model fields that we introduce are pretty simple. 

... OK, in an effort to figure out what I'm missing about testing models, I skimmed the earlier chapters of the book again, and found this example: 
```
from lists.models import Item
[...]

class ItemModelTest(TestCase):

    def test_saving_and_retrieving_items(self):
        first_item = Item()
        first_item.text = 'The first (ever) list item'
        first_item.save()

        second_item = Item()
        second_item.text = 'Item the second'
        second_item.save()

[...]

PR #184, opened 2020-09-12
[Issue 179] Adding a test for the osprojects model

lpatmo:
## What type of PR is this? (check all applicable)

- [ ] 🐛 Bug Fix
- [ ] ✨ Feature (_**e.g.** API change, enhancement, addition, etc._)
- [ ] ♻️ Refactor
- [ ] 📝 Documentation
- [ ] 🔖 Release
- [x] 🚩 Other (**concerning:**  tests)

## Context

Tests the `osprojects` model we created! Two todos: 

[x] Cleared up this issue: ```osprojects/tests.py::test_osprojects_model
  /usr/local/lib/python3.7/site-packages/django/db/models/fields/__init__.py:1363: RuntimeWarning: DateTimeField OSProjects.modified received a naive datetime (2020-05-31 00:00:00) while time zone support is active.
    RuntimeWarning)``` by using `datetime.now()`
[ ] Add more asserts (maybe?)

## Other Related Tickets & Documents (as needed)

#179 

## Implementation Details
[x] Created a factories.py file to generate new osprojects
[x] Tested that a newly created `osproject` had the relevant fields
[x] Ran `docker-compose run --rm app pytest` and confirmed that tests passed

## New Libraries/Dependancies Introduced (Fill out as needed)
N/A

If you have added libraries or other dependancies, please list them (and links to their repos) below:

- [ ] I've updated `requirements.txt`

## Any new migration files added?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed

## Did you add tests?
**_Code added or changed without test coverage or good reason for exemption won't be approved._**

Yes, but could use feedback on these tests!

- [x] 👍 yes
- [ ] 🙋 no, because I need help
- [ ] 🙅 no, because they aren't needed

## Did you add documentation?

- [ ] 📜 readme.md
- [ ] 📜 contributing.md
- [ ] 📜 wiki entry
- [ ] 🙋 I'd like someone to help write documentation, and will file a new issue for it
- [x] 🙅 no documentation needed

### DIFF
--- a/project/osprojects/models.py
+++ b/project/osprojects/models.py
@@ -7,6 +7,7 @@ from django.contrib.auth import get_user_model
 from tagging.managers import CustomTaggableManager
 from tagging.models import CustomTag, TaggedItems
 
+
 def get_sentinel_user():
     return get_user_model().objects.get_or_create(username='deleted')[0]
 
--- a/project/osprojects/tests.py
+++ b/project/osprojects/tests.py
@@ -1,3 +1,11 @@
-from django.test import TestCase
+import pytest
+from .factories import OSProjectsFactory
 
-# Create your tests here.
+
+@pytest.mark.django_db
+def test_osprojects_model():
+    """ Test OSProjects model """
+    # create OSProjects model instance
+    osproject = OSProjectsFactory(title="Buddybot", guid="cc63f918-d515-11ea-956f-0242ac130002")
+    assert osproject.title == "Buddybot"
+    assert osproject.guid == "cc63f918-d515-11ea-956f-0242ac130002"
--- a/project/osprojects/factories.py
+++ b/project/osprojects/factories.py
@@ -1,0 +1,25 @@
+import uuid
+from random import choice
+from factory import DjangoModelFactory, Faker, fuzzy, LazyAttribute, SubFactory
+import datetime
+from users.factories import UserFactory
+from .models import OSProjects
+
+
+class OSProjectsFactory(DjangoModelFactory):
+    """
+        Define OSProjects Factory
+    """
+    guid = LazyAttribute(lambda obj: uuid.uuid1())
+    title = Faker("sentence")
+    project_creator = Faker("name")
+    description = Faker("text")
+    url = Faker("uri")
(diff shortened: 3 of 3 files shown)


project/osprojects/models.py
```diff
@@ -7,6 +7,7 @@
 from tagging.managers import CustomTaggableManager
 from tagging.models import CustomTag, TaggedItems
 
+
```
lpatmo, line 10: Note: my linter caused this extra line, but not sure why it's appearing 🤔

project/osprojects/tests.py
```diff
@@ -1,3 +1,11 @@
...
+from .factories import OSProjectsFactory
 
-# Create your tests here.
+
+@pytest.mark.django_db
+def test_osprojects_model():
+    """ Test OSProjects model """
+    # create OSProjects model instance
+    osproject = OSProjectsFactory(title="Buddybot", guid="cc63f918-d515-11ea-956f-0242ac130002")
+    assert osproject.title == "Buddybot"
```
lpatmo, line 10: Note: was thinking I could add more asserts here, but was wondering if this might also be fine

project/osprojects/models.py
BethanyG: Apologies for the brain fart and not reviewing this sooner!  I think your linter is enforcing some sort of PEP 8 thing where imports are separated from code by two lines?  Dunno.  Weird.

project/osprojects/tests.py
BethanyG: Up to you.  I usually go for an entire record -- or even two, then pick up to four things to assert.  But even one proves it was inserted into the DB, which is what we are going for here....so.....

BethanyG (approved):
I think this works.  You could get more elaborate, or even update data and check to make sure it works.  Although that's getting into re-testing the ORM territory.  I think this is probably "good enough" for a first pass at a model test for `OSProjects`.  We can revisit as we get more of the endpoint done.

lpatmo:
Thank you for reviewing, @BethanyG!! Will merge this in... and as you said, revisit/update as we get more of the endpoint done.

